PR #45, opened 2015-06-26
Display a helpful banner after installing the analytics plugin

justinshreve:
I mentioned this idea to @claudiosmweb the other day, but displaying some additional information (that may have been missed in the readme) would help users understand some of the requirements of the plugin and might reduce some of the initial "it's broken"/"my analytics are not showing" support requests. Think of it as a super simple onboarding message :).

The notice can be permanently dismissed by the user, only displays on WooCommerce settings pages, and will also disappear when the GA ID is set (so it should not show for upgraded users unless they have never configured GA before).

Here is what it looks like:
preview of info banner

### DIFF
--- a/includes/class-wc-google-analytics-info-banner.php
+++ b/includes/class-wc-google-analytics-info-banner.php
@@ -29,7 +29,7 @@ class WC_Google_Analytics_Info_Banner {
 	public function __construct( $dismissed = false, $ga_id = '' ) {
 		$this->is_dismissed = (bool) $dismissed;
 		if ( ! empty( $ga_id ) ) {
-			//$this->is_dismissed = true;
+			$this->is_dismissed = true;
 		}
 
 		// Don't bother setting anything else up if we are not going to show the notice
--- a/includes/class-wc-google-analytics-info-banner.php
+++ b/includes/class-wc-google-analytics-info-banner.php
@@ -60,6 +60,7 @@ class WC_Google_Analytics_Info_Banner {
 			$configure,
 			__( 'After configuring the plugin, please give Google Analytics 24 hours to start displaying results.', 'woocommerce-google-analytics-integration' ),
 			__( 'For transaction tracking to properly work, you will need to use a payment gateway that redirects the customer back to a WooCommerce order received/thank you page.', 'woocommerce-google-analytics-integration' ),
+			__( 'You must enable Ecommerce reporting in Google Analytics. <a href="https://support.google.com/analytics/answer/1009612?hl=en#Enable">See here for more information</a>.')
 		);
 
 		// Display the message..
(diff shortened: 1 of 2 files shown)


claudiosmweb:
@BFTrick any thoughts?

BFTrick:
I have mixed feelings on admin notices. I know a lot of people ignore them based on my time in support. You would login and they'd have admin noticed permanently fixed to the top of the page and they don't bother to get rid of.

What about redirecting to the WC GA settings page after activation?

mikejolley:
@BFTrick @justinshreve I feel the main objection to banners was due to the amount WC core added on install previously. There were ~3/4. I think in general, this will be semi-ok.

I would however tweak this to be less of a wall of text:

1. Show a setup prompt (no auto redirect) just saying you need to setup the extension to connect to GA. Button CTA links to settings page.
2. Saving settings success message should state, in bold, wait 24 hours and about the gateways

Related, @justinshreve is there a GA method to trigger tracking from the server side, for gateways which don't play nice? If so, the payment hooks could possibly be used instead.

justinshreve:
@mikejolley It looks like there is now a way to do this server side https://developers.google.com/analytics/devguides/collection/protocol/v1/ (also https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide#ecom and https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide#enhancedecom). It does require Universal Analytics to be configured (and https://developers.google.com/analytics/devguides/collection/protocol/policy makes it seem like we might need to offer a user facing opt out?).

One thing I worry about is having this plugin be confusing to setup.... ie you set it up and have all these checkboxes. You have standard vs universal tracking. You have eCommerce vs Enhanced eCommerce... now potentially a server side vs client side tracking option. Not sure what the solution is since we have to support the other use cases for backwards compatibility... it just makes the plugin kind of confusing to use out of the box.

BFTrick:
> One thing I worry about is having this plugin be confusing to setup.... ie you set it up and have all these checkboxes. You have standard vs universal tracking. You have eCommerce vs Enhanced eCommerce... now potentially a server side vs client side tracking option. Not sure what the solution is since we have to support the other use cases for backwards compatibility... it just makes the plugin kind of confusing to use out of the box.

Totally agree with it being confusing. I don't think we should support server side. That's going to go into the Google Analytics Pro extension and it's similar to Yoast's premium product as well. I think it's fine that this is only client side.

@justinshreve can we add some tracking for this extension? If someone has opted into WC tracking maybe we can pass a few more points of data? Ex. what type of analytics they're using (universal or legacy; enhanced ecommerce or regular, etc). Then down the road if we see all new installs using universal we can make that the default and hide the option.

justinshreve:
@BFTrick I think that's a great idea. Tracked it at #46.

justinshreve:
I made the two adjustments that Mike mentioned above.

PR #47, opened 2015-06-29
Refactor of JS tracking code

justinshreve:
This PR does not add any new functionality. It's a larger refactor of the code (which I mentioned in Slack last week) to make future improvements and changes easier (such as enhanced ecommerce tracking). We can probably make some more improvements, but I think it's a start and an improvement over the current code.

The main thing I focused on was reducing code duplication. For example, despite there being differences in ga.js vs universal tracking, there was still some code (like `$logged_in`) used by both. eCommerce vs standard tracking also duplicated a lot of code. My solution to this was to create a new helper class with the different JS generating methods (loading analytics, transaction, add item, etc) so they could be called from the controller methods in the main class. I also fixed up some coding standards issues, removed some extra unused code, and made some of the comments more accurate.

We'll probably want to do some additional testing on this (+ other changes like enhanced ecommerce) before doing a full release. But using the GA developer plugin (and viewing results in the dashboard) things are getting tracked accurately :).

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -293,5 +293,28 @@ class WC_Google_Analytics_JS {
 		return $code;
 	}
 
+	/**
+	 * Add to cart
+	 *
+	 * @param array $parameters associative array of _trackEvent parameters
+	 * @param string $selector jQuery selector for binding click event
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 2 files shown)


mikejolley:
We can do a beta test when enhanced stuff is ready - I'm certain there will be many volunteers and this repo is public :)

Issue #51, opened 2015-07-03
Group "Enhanced eCommerce" options


mikejolley:
Rather than show these checkboxes with the description:

"Requires Enhanced eCommerce."

I think it would be good to instead hide that description and use JS to only reveal the options when enhanced ecommerce is checked. Less options = Less scary settings pages.

lkraav:
Not sure if this is worth a new issue, but Enhanced E-commerce should be hidden if "Universal Analytics" is unchecked. As far as I can tell based on [link](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce), these states are mutually exclusive.

justinshreve:
Good point. I opened up a new issue for it and will work on that tomorrow. Thanks!

PR #59, opened 2015-08-04
Hide all of the enhanced eCommerce settings if Universal Analytics is disabled

### LINKED ISSUE #58: Hide enhanced eCommerce settings if universal analytics is unchecked
Enhanced eCommerce should be hidden if "Universal Analytics" is unchecked since universal analytics is required.

https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce
https://github.com/woothemes/woocommerce-google-analytics-integration/issues/51#issuecomment-127370803

justinshreve:
Fixes #58.

Since UA is required for enhanced eCommerce to work (and we won't load the enhanced eCommerce code without that setting also being on) we will hide enhanced settings when US is turned off.


justinshreve:
See #51

PR #92, opened 2017-04-28
Add an empty product check.

### LINKED ISSUE #81: Fatal Error when using Enhanced eCommerce Product Detail Views and WooCommerce [product] shortcodes
If you have the "Product Detail Views" "Enhanced eCommerce" setting enabled, and you use a `[product id="x"]` shortcode on a WordPress page, a fatal error occurs:

`Call to a member function get_sku() on a non-object in wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-google-analytics-js.php on line 445`

Also reported here: https://wordpress.org/support/topic/bug-conflict?replies=1

bor0:
Fixes #81.

-------------------

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -455,7 +455,7 @@ class WC_Google_Analytics_JS {
 	 * Tracks a product detail view
 	 */
 	function product_detail( $product ) {
-		if ( ! empty( $product ) ) {
+		if ( empty( $product ) ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


includes/class-wc-google-analytics-js.php
```diff
@@ -455,6 +455,10 @@ function remove_from_cart() {
 	 * Tracks a product detail view
 	 */
 	function product_detail( $product ) {
+		if ( ! empty( $product ) ) {
```
gedex: shouldn't this check if product is empty? because the following lines use the product instance.

PR #109, opened 2018-01-30
Add filter to optionally disable all tracking

### LINKED ISSUE #108: Feature Request: Hook to Disable Tracking
What do we think about adding a hook so other plugins can control whether the tracking is enabled in certain situations?

My use case is that I want to exclude my traffic from tracking. Unfortunately, I have several dynamic IPs, so I have written a plugin to periodically update my IPs. Now, I just need a way to disable tracking on traffic from those IPs.

I'm thinking we could simply wrap [this line](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/1.4.3/includes/class-wc-google-analytics.php#L349) with `apply_filters( ...`. I would be happy to put together a pull request if you would like.

dominic-p:
This adds a simple filter to the `disable_tracking` method and then removes some duplicate code in the `tracking_code_display` method to make it work. Feedback is welcome.

Fixes #108.

- Add a filter so other plugins can disable analytics tracking on demand

-------------------
- [x] Did you make changes, or create a **new .js file**? Make sure you ran `grunt` before creating this Pull Request.

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -281,7 +281,7 @@ class WC_Google_Analytics extends WC_Integration {
 		global $wp;
 		$display_ecommerce_tracking = false;
 
-		if ( is_admin() || current_user_can( 'manage_options' ) || ! $this->ga_id ) {
+		if ( $this->disable_tracking( 'all' ) ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


bor0:
Thank you for your contribution!

dominic-p:
Of course. Thanks for merging it. :)

PR #125, opened 2018-06-29
Fix fatal error on Thank You page

### LINKED ISSUE #124: Fatal error on Thank You page with invalid order ID
#### What I expected

A blank order received page.

#### What happened instead

Fatal PHP error.

#### Steps to reproduce the issue

Visit the order received page with an order ID that doesn't exist. For example:

http://www.example.com/checkout/order-received/999999/

I found the code that is responsible, and a PR is incoming.

-------------------

- [ ] Issue assigned to next milestone.
- [ ] Issue assigned a priority (will be assessed by maintainers).

dominic-p:
Fixes #124.

If we have an invalid order ID, we wind up throwing a fatal PHP error. This seems to solve the problem.

- Use `wc_get_order()` instead of `new WC_Order()`
- Validate the order object before we try to use it.

-------------------

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -321,7 +321,10 @@ class WC_Google_Analytics extends WC_Integration {
 	 */
 	protected function get_ecommerce_tracking_code( $order_id ) {
 		// Get the order and output tracking code
-		$order = new WC_Order( $order_id );
+		$order = wc_get_order( $order_id );
+
+		// Make sure we have a valid order object
+		if ( ! $order || is_wp_error( $order ) ) return "";
 
 		$code = WC_Google_Analytics_JS::get_instance()->load_analytics( $order );
 		$code .= WC_Google_Analytics_JS::get_instance()->add_transaction( $order );
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -320,24 +320,26 @@ class WC_Google_Analytics extends WC_Integration {
 	 * @param int $order_id
 	 */
 	protected function get_ecommerce_tracking_code( $order_id ) {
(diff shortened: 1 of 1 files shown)


includes/class-wc-google-analytics.php
```diff
@@ -321,7 +321,10 @@ protected function get_standard_tracking_code() {
 	 */
 	protected function get_ecommerce_tracking_code( $order_id ) {
 		// Get the order and output tracking code
-		$order = new WC_Order( $order_id );
+		$order = wc_get_order( $order_id );
+
+		// Make sure we have a valid order object
+		if ( ! $order || is_wp_error( $order ) ) return "";
```
roykho: Thanks! Some notes on this. Errors are already handled for us so you can simplify to `if ( ! $order ) {...}` And mind coding standards such as period at the end of the comment and no inline condition statements like that. And use single quotes instead of double quotes.

dominic-p: Thanks for the feedback. I'm happy to make all of those changes, but I wanted to double check on some things first.

1. If the `$order` variable was indeed an instance of WP_Error, wouldn't the simple `! $order` miss it? Or are you saying that `wc_get_order()` will never return a `WP_Error` object?

2. I tried to match the code style of the rest of the function regarding periods at the end of the comment and quote style. I can definitely change that, but then it wouldn't match what's there. Please confirm.

roykho: Exception is handler up the chain and when caught will return a false to the caller. Yeah so the existing code may have been written long ago and possibly by different people so any new code we add it must follow the WP coding standards.

dominic-p:
No problem. Thanks for merging it. I kind of thought you meant that, but the mixed styles just grated on me too much. :) I'll only format my code in the future.

PR #133, opened 2018-11-15
Use wc_enqueue_js instead of echo for listing_click

### LINKED ISSUE #132: Add to cart JavaScript is included right after the add to cart link
In the function `listing_click`, the JavaScript is echo'd to the page instead of being enqueued. This can cause some theme compatibility issues. In this case a problem with the JavaScript showing up in tooltips, see: https://wordpress.org/support/topic/code-leak-on-tooltips/

I'd suggest to use `wc_enqueue_js` instead, which is the same that is done for the `listing_impression` function.

mikkamp:
Fixes #132

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -155,26 +155,22 @@ class WC_Google_Analytics_JS {
 			$list = "Product List";
 		}
 
-		echo( "
-			<script>
-			(function($) {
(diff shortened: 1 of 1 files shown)


mikkamp:
@bor0 or @c-shultz 
This has been approved, but is there anything else that needs to be done or can this be merged so this doesn't get missed when a new release happens?

PR #134, opened 2018-11-26
Update class-wc-google-analytics.php

miguelfspinto:
Remove is_admin validation because it is avoiding GA events on AJAX requests

Fixes # .

-

-------------------

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -350,7 +350,7 @@ class WC_Google_Analytics extends WC_Integration {
 	 * @return bool True if tracking for a certain setting is disabled
 	 */
 	private function disable_tracking( $type ) {
-		if ( is_admin() || current_user_can( 'manage_options' ) || ( ! $this->ga_id ) || 'no' === $type || apply_filters( 'woocommerce_ga_disable_tracking', false, $type ) ) {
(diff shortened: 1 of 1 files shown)


includes/class-wc-google-analytics.php
```diff
@@ -350,7 +350,7 @@ protected function get_ecommerce_tracking_code( $order_id ) {
 	 * @return bool True if tracking for a certain setting is disabled
 	 */
 	private function disable_tracking( $type ) {
-		if ( is_admin() || current_user_can( 'manage_options' ) || ( ! $this->ga_id ) || 'no' === $type || apply_filters( 'woocommerce_ga_disable_tracking', false, $type ) ) {
+		if ( current_user_can( 'manage_options' ) || ( ! $this->ga_id ) || 'no' === $type || apply_filters( 'woocommerce_ga_disable_tracking', false, $type ) ) {
```
roykho, line 353: Instead of removing that perhaps check if its doing AJAX? `if ( ( ! wp_doing_ajax() && is_admin() ) || ...`

mikkamp:
In addition to it being better to explicitly check for AJAX requests. WooCommerce core sends all it's AJAX requests through a frontend URL with the parameter `?wc_ajax=<action>`
When a request like this is sent `is_admin` does not return true, so I don't think we should disable this check.

If there are certain plugins still sending AJAX requests through a backend URL, which need to be tracked, then the filter `woocommerce_ga_disable_tracking` is available to override this behaviour.

Based on this information I'm going to close this pull request.

PR #135, opened 2018-11-26
Update class-wc-google-analytics.php

miguelfspinto:
add filter to add more fields in product object like brand and category

my recomments is instead of using string concatenation to create a object, we should use php array, then apply the filter in order to add more entries in the array and then, concatenate with json_encode inside the JS method. 
With this, is very easy to add new items to the object like brand and category.

Fixes # .

-

-------------------


mikkamp:
Thanks for the PR, although we decided to look into a more standardized approach for adding hooks and filters, so we will be closing this PR and tracking the needed changes through an issue. See: https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/171#issuecomment-705408207

The code changes look like a nice solution, although we generally want to stick to using standard names for the filters throughout the plugin.

PR #140, opened 2019-02-19
Make analytics.js src filterable

Dan0sz:
This way other plugins can override the source, for example by a local copy (e.g. provided by CAOS for Analytics).

- I've been getting many requests for this on the WP Support Forums and by e-mail. So I decided to make a pull request. Adding a filter the the analytics.js source URL will make it easy for plugins to alter the source of the analytics.js file. Some people want to host it locally, to up their Pagespeed score. My plugin does that, and keeps it updated using wp-cron. If you'd implement this, I could create an option to make my plugin compatible with yours! :)

-------------------
- [x] Did you make changes, or create a **new .js file**? Make sure you ran `grunt` before creating this Pull Request.


miketlp:
Yes, please consider Dan0sz's request. I would also like to see this made so that I can locally host Google Analytics on our website.

bor0:
Thank you for your contribution!

PR #162, opened 2020-03-18
Change wc_goole_analytics_send_pageview fiter name to wc_google_analytics_send_pageview.

budzanowski:
Update typo in filter name while providing soft deprecation with deprecation notice.

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -204,7 +204,7 @@ class WC_Google_Analytics_JS {
 	}
 
 	/**
-	 * This was created to fix public facing api typo in a filter nam
+	 * This was created to fix public facing api typo in a filter name
 	 * and inform about the deprecation.
 	 */
 	public static function universal_analytics_footer_filter( $send_pageview ) {
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -71,6 +71,7 @@ class WC_Google_Analytics_JS {
 	 */
 	public static function load_analytics( $order = false ) {
 		$logged_in = is_user_logged_in() ? 'yes' : 'no';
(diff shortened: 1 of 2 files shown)


includes/class-wc-google-analytics-js.php
```diff
@@ -197,11 +198,24 @@ public static function classic_analytics_footer() {
...
+	 * This was created to fix public facing api typo in a filter nam
+	 * and inform about the deprecation.
+	 */
+	public static function universal_analytics_footer_filter( $send_pageview ) {
+		if ( has_filter( 'wc_goole_analytics_send_pageview' ) ) {
+			$log_string  = "The wc_goole_analytics_send_pageview filter is deprecated since version 1.4.20.";
+			$log_string .= "Replace with wc_google_analytics_send_pageview";
+			error_log( $log_string );
+		}
+		return apply_filters( 'wc_goole_analytics_send_pageview', $send_pageview );
```
bor0: I just found out about [`apply_filters_deprecated`](https://developer.wordpress.org/reference/functions/apply_filters_deprecated/). Its definition seems to be just what we need:

```php
function apply_filters_deprecated( $tag, $args, $version, $replacement = false, $message = null ) {
    if ( ! has_filter( $tag ) ) {
        return $args[0];
    }
 
    _deprecated_hook( $tag, $version, $replacement, $message );
 
    return apply_filters_ref_array( $tag, $args );
}
```

This will further simplify the code:

```suggestion
		$send_pageview = apply_filters_deprecated( 'wc_goole_analytics_send_pageview', array( $send_pageview ), '1.4.20', 'wc_goole_analytics_send_pageview' );
		return apply_filters( 'wc_goole_analytics_send_pageview', $send_pageview );
```

budzanowski: Good idea but I think that we just need to return the `apply_filters_deprecated`. Please see 8716b604e14c6344b286a5529e4cbee6880ea930

PR #165, opened 2020-05-07
Use local store of orders to avoid duplicate transaction logging

### LINKED ISSUE #120: GA reporting duplicate sales
1163667-zen
1423824-zen

Google Analytics (https://woocommerce.com/products/woocommerce-google-analytics/) is recording duplicate sales from WooCommerce. Here's an example: https://snag.gy/PXLQUZ.jpg. The quantity should be "1" and amount "$39.95."

This seems to be happening randomly, and one possible theory that came up was that the duplicate logging was occurring when customers visit the Order Received page/thank you page twice: https://www.lunametrics.com/blog/2014/07/07/duplicate-transactions/

Is this the case with our extension? Or do we have code in place to prevent logging users from triggering the script on the confirmation page upon return to it? 

Any other thoughts at what could be causing this?

@mattyza - would you be able to provide any insight on this for me? TIA!

tc33:
### All Submissions:

When transaction events are fired order numbers are saved to a localStorage item, and transactions are only fired if the localStorage item does not already contain that order number. This avoids duplicate transactions being tracked for the same order, which is possible when the thank-you page reloads without causing a server request. Duplicate transactions could still be possible if the user manipulates the localStorage item, but it should be much less common.

Closes #120 

1. Complete an order to confirm the transaction event still fires on the thank-you page
2. Manually remove the _ga_tracked post meta value to remove the server side protection
3. Reload the thank-you page to confirm transaction event does not fire a second time
4. The ga_orders localStorage item can be checked manually in developer tools

The issue of the thank-you page being reloaded without hitting the server happens when tabs are restored on some devices. You could perform testing in that scenario, but I couldn't confirm exactly which devices/browsers behave like this, except that iOS is most prevalent in the analytics.

### Changelog entry

Fix - Prevent duplicate transactions from browser reload

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -288,13 +288,23 @@ class WC_Google_Analytics_JS {
 	function add_transaction( $order ) {
 		if ( 'yes' == self::get( 'ga_use_universal_analytics' ) ) {
 			if ( 'yes' === self::get( 'ga_enhanced_ecommerce_tracking_enabled' ) ) {
-				return self::add_transaction_enhanced( $order );
+				$transaction_code = self::add_transaction_enhanced( $order );
 			} else {
-				return self::add_transaction_universal( $order );
+				$transaction_code = self::add_transaction_universal( $order );
 			}
 		} else {
-			return self::add_transaction_classic( $order );
+			$transaction_code = self::add_transaction_classic( $order );
 		}
+
+		// Wrap the transaction in a check against localStorage to avoid duplicate firing if page is reloaded without hitting server
+		$code = "var gaOrders = localStorage.getItem('ga-orders');";
+		$code .= "gaOrders = (gaOrders !== null) ? gaOrders : '';";
+		$code .= "if (gaOrders.split(';').indexOf('" . esc_js( $order->get_order_number() ) . "') === -1) {";
+		$code .= $transaction_code;
+		$code .= "localStorage.setItem('ga-orders', gaOrders + '" . esc_js( $order->get_order_number() ) . "');";
+		$code .= "}";
+
+		return $code;
 	}
 
 	/**


includes/class-wc-google-analytics-js.php
```diff
@@ -288,13 +288,23 @@ public static function load_analytics_universal( $logged_in ) {
...
+			$transaction_code = self::add_transaction_classic( $order );
 		}
+
+		// Wrap the transaction in a check against localStorage to avoid duplicate firing if page is reloaded without hitting server
+		$code = "var gaOrders = localStorage.getItem('ga-orders');";
+		$code .= "gaOrders = (gaOrders !== null) ? gaOrders : '';";
+		$code .= "if (gaOrders.split(';').indexOf('" . esc_js( $order->get_order_number() ) . "') === -1) {";
+		$code .= $transaction_code;
+		$code .= "localStorage.setItem('ga-orders', gaOrders + '" . esc_js( $order->get_order_number() ) . "');";
+		$code .= "}";
```
lkraav, line 305: May I suggest a readability upgrade here

```php
$code = sprintf( <<<JS
  var gaOrders = localStorage.getItem('ga-orders');
  gaOrders = (gaOrders !== null) ? gaOrders : '';
  if (gaOrders.split(';').indexOf('%1$s') === -1) {
    %2$s;

    localStorage.setItem('ga-orders', gaOrders + '%1$s');
  }
JS
, esc_js( $order->get_order_number() ), $transaction_code );
```

tc33: The way it's been done is more consistent with the existing code base. I do agree it could be improved though. I'd suggest making that sort of refactor in a separate pull request to change it across the whole plugin at the same time.

brad82:
I have been using this PR content in production for a month and can confirm it has a positive effect on reducing double transactions from iOS (/mobile) devices.

mikkamp:
Closing as this has been resolved in #170

Issue #171, opened 2020-08-21
Filter needed to modify transaction data


ajayghaghretiya-multidots:
#### What I expected

We have modification data for the product on our website and we need to track all the custom information on Google Analytics Tools. We are using this plugin. And it did very much help with that. 

But It would be really great we have some control on data when calling any transaction event. And we all know the modification in the plugin is not a good thing.

So If we have some filters to modify the data from the theme that will be very helpful.

Also, GA supporting more fields https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#product-data but we don't have filter to add the fields & it's data.

#### What happened instead
Don't have any filter to modify the enhanced purchase transaction data of the Order's item.

We have limited static fields for the enhanced purchase transaction in the path **_woocommerce-google-analytics-integration/includes/class-wc-google-analytics-js.php_**

                `function add_item_enhanced( $order, $item ) {
		$_product = version_compare( WC_VERSION, '3.0', '<' ) ? $order->get_product_from_item( $item ) : $item- 
                >get_product();
		$variant  = self::product_get_variant_line( $_product );

		$code = "" . self::tracker_var() . "( 'ec:addProduct', {";
		$code .= "'id': '" . esc_js( $_product->get_sku() ? $_product->get_sku() : $_product->get_id() ) . "',";
		$code .= "'name': '" . esc_js( $item['name'] ) . "',";
		$code .= "'category': " . self::product_get_category_line( $_product );

		if ( '' !== $variant ) {
			$code .= "'variant': " . $variant;
		}

		$code .= "'price': '" . esc_js( $order->get_item_total( $item ) ) . "',";
		$code .= "'quantity': '" . esc_js( $item['qty'] ) . "'";
		$code .= "});";

		return $code;
	} `

#### Steps to reproduce the issue
1. If you have brand specified on the site via product tag or any custom field. It will not log to GA with the field while GA supporting that field.

-------------------

- [ ] Issue assigned to next milestone.
- [ ] Issue assigned a priority (will be assessed by maintainers).

mikkamp:
Thanks for providing a suggested PR. It's a nice solution but we are going to need a more standardized approach for adding hooks to the various snippets which are output.

[...]

PR #172, opened 2020-08-21
fixed the static parameters of the google analytic order item enhanced

ajayghaghretiya-multidots:
#171 Add filter to update/modify the the data of the purchase transaction enhanced item. Can also add more item's fields with this filter like brand etc
https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#product-data

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -442,18 +442,19 @@ class WC_Google_Analytics_JS {
 		$_product = version_compare( WC_VERSION, '3.0', '<' ) ? $order->get_product_from_item( $item ) : $item->get_product();
 		$variant  = self::product_get_variant_line( $_product );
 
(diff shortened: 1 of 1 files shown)


mikkamp:
Thanks for the PR, although we decided to look into a more standardized approach for adding hooks and filters, so we will be closing this PR and tracking the needed changes through the issue. See: https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/171#issuecomment-705408207

The code changes look like a nice solution, although we generally recommend to use `wp_json_encode` instead of `json_encode`. The option `JSON_UNESCAPED_UNICODE` is also needed, to prevent it from converting unicode characters. While the creation of an array makes it easy to filter, it won't work in all cases (for example where the values are set through jQuery), so for now we are going to look into what will be an overall best solution.

PR #197, opened 2021-08-30
Avoid unneccessarily running the completed orders query

### LINKED ISSUE #163: Google Analytics Pro notices option will not be set if a site already has 100 completed orders on plugin activation
`woocommerce_google_analytics_pro_notice_shown`  option will never get set is the site where the plugin is activated on already has 100 completed orders. Then the sad query will keep being run.

ArtskydJ:
### All Submissions:

The `woocommerce_google_analytics_pro_notice_shown` now gets set to `true` when there are over 100 completed orders, so that the query counting the orders will not be executed again.

Closes #163.

### Other information:

### Changelog entry

> Tweak - Avoid unnecessary completed orders queries.

### DIFF
--- a/woocommerce-google-analytics-integration.php
+++ b/woocommerce-google-analytics-integration.php
@@ -134,7 +134,12 @@ if ( ! class_exists( 'WC_Google_Analytics_Integration' ) ) {
 			$completed_orders = wc_orders_count( 'completed' );
 
 			// Only show the notice if there are 10 <= completed orders <= 100.
-			if ( ! ( 10 <= $completed_orders && $completed_orders <= 100 ) ) {
+			$too_few_orders_to_show_the_notice = $completed_orders < 10;
+			$too_many_orders_to_show_the_notice = $completed_orders > 100;
+			if ( $too_many_orders_to_show_the_notice ) {
+				update_option( 'woocommerce_google_analytics_pro_notice_shown', true );
+			}
+			if ( $too_few_orders_to_show_the_notice || $too_many_orders_to_show_the_notice ) {
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


woocommerce-google-analytics-integration.php
```diff
@@ -134,7 +134,12 @@ public function show_ga_pro_notices() {
 			$completed_orders = wc_orders_count( 'completed' );
 
 			// Only show the notice if there are 10 <= completed orders <= 100.
-			if ( ! ( 10 <= $completed_orders && $completed_orders <= 100 ) ) {
+			$too_few_orders_to_show_the_notice = $completed_orders < 10;
+			$too_many_orders_to_show_the_notice = $completed_orders > 100;
```
layoutd:
```suggestion
			$too_few_orders_to_show_the_notice  = $completed_orders < 10;
			$too_many_orders_to_show_the_notice = $completed_orders > 100;
```
Align variable assignments.

layoutd (changes requested):
Thanks for this PR, @Artskydj. 👍🏻  Just one small code style suggestion.

We've also had some discussion about the action itself: could you change it from `init` to `admin_init` (so the notice logic isn't run on every single public page load) before we merge the PR?
https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/5e54e59ae19cf8be949ff6dbcca142db01bfb6f3/woocommerce-google-analytics-integration.php#L43

Once those two changes are made, we can merge the PR, and and include the changes in next week's release!

woocommerce-google-analytics-integration.php
ArtskydJ: Fixed in c9449d8

ArtskydJ:
Thanks!  I've updated `init` ⏩ `admin_init`, and fixed the unaligned whitespace.

layoutd (approved):
Looks good, thanks for the changes @ArtskydJ! 

I'll merge the PR, and the changes should be included in release 1.5.3 early next week.

PR #203, opened 2021-12-06
Add support for cross domain tracking features

workeffortwaste:
### All Submissions:

Adds support for Google Analytics cross domain tracking features (automatic link decoration and accepting linker parameters) for both Global Site Tag and Universal Analytics.

1. There is a new section in the settings with the cross domain options. Adding in a domain such as example.com to the cross domain text field will automatically decorate any outbound links to example.com with a tracking parameter upon click. 
2. To test automatic acceptance of linker parameters the Google Analytics Debugger Chrome extension will show when incoming linker parameters (?_ga) in the query string have been used to set the Google Analytics Client ID.

### Other information:

### Changelog entry

> Add - Support for Google Analytics cross domain tracking features.

### DIFF
--- a/includes/class-wc-google-analytics-js.php
+++ b/includes/class-wc-google-analytics-js.php
@@ -245,9 +245,9 @@ class WC_Google_Analytics_JS extends WC_Abstract_Google_Analytics_JS {
 		$ga_id = self::get( 'ga_id' );
 
 		if( 'yes' === self::get( 'ga_linker_allow_incoming_enabled' ) ) {
-			$ga_snippet_create       = self::tracker_var() . "( 'create', '" . esc_js( $ga_id ) . "', '" . $set_domain_name . "', { allowLinker: true });";
+			$ga_snippet_create = self::tracker_var() . "( 'create', '" . esc_js( $ga_id ) . "', '" . $set_domain_name . "', { allowLinker: true });";
 		} else {
-			$ga_snippet_create       = self::tracker_var() . "( 'create', '" . esc_js( $ga_id ) . "', '" . $set_domain_name . "' );";
+			$ga_snippet_create = self::tracker_var() . "( 'create', '" . esc_js( $ga_id ) . "', '" . $set_domain_name . "' );";
 		}
 
 		if ( ! empty( self::DEVELOPER_ID ) ) {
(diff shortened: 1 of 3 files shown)


findingsimple:
Nice! Thanks for the contribution @workeffortwaste 

Could you link to or give us a bit more context on the functionality? We'll need to get up to speed on whether we should introduce this or not.

workeffortwaste:
Sure, no problem. 

Here's Google's documentation for Cross Domain Measurement: https://developers.google.com/analytics/devguides/collection/gtagjs/cross-domain
https://support.google.com/analytics/answer/1033876

Essentially it allows you to continue a Google Analytics session across two different domains, an example of why this is useful would be if a website has their primary site on example.com but has their shop on exampleshop.com.

With the same Analytics account active on each site and cross domain linking in place any clicks through to the exampleshop.com from example.com would be decorated with a linking parameter in the query string. This would be picked up exampleshop.com and the Analytics session would be continued, rather than a fresh session started.

Where this becomes really useful is if you're running Google Ads to example.com. 

Without cross domain linking all the sales on the shop site that originated from ad clicks to example.com wouldn't be correctly attributed in Google Analytics, as it wouldn't be able to link the two sessions together, and would instead show as direct traffic. 

With cross domain linking in place Google Analytics will be able to attribute all those sales correctly back to their original ad clicks and campaigns (or other source/medium).

Hopefully this makes sense.

mikkamp:
@workeffortwaste Thanks again for the contribution, and sorry for the late reply. We did have a look at the contribution and think it would be a useful addition, although we did have a couple small comments.

One was to either use only one option, or alternatively hide the second option field conditionally (you can see an example of how fields are toggled [here](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/trunk/assets/js/admin-ga-settings.js)). This would prevent the options screen from getting too many options which aren't relative for every site.

Another thing we noticed is that there are several spacing issues which aren't following the [WordPress coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/). We realize not all the code in the extension follows the standards, but we do try to the standards for any new code we add.

workeffortwaste:
@mikkamp I've fixed the spacing issues to conform to the WordPress coding standards.

With regards to having only a single option or hiding one until the other is used, I'd argue having both visible is a requirement as there are instances where you would want one option but not the other.

Scenario 1: You want to cross link with an external domain, but there are no incoming links from that domain so you don't require auto linker to be enabled.

_Only a comma separated list of domains is required._

Scenario 2: You want to accept incoming links with linker parameters from an external domain, but you have no outgoing links to that domain.

_Only accepting incoming linker parameters is required._

mikkamp (approved):
Thanks for the clarification. I did some more testing and it's outputting all the information as expected. I just added a few more small phpcs fixes. I also noticed that the tracking data was swapped around so it was outputting `yes`/`no` for the domain names instead of incoming links being enabled.

PR #206, opened 2021-12-27
Confirm order key when injecting transaction tracking code

layoutd:
### All Submissions:

This pull request ensures that a valid order key is present on the "Order Received" page before inserting the GA tracking code.

1. Enable a payment method in WooCommerce - either credit card processing, or Cash on Delivery.
1. Enable WooCommerce Google Analytics Integration and configure the settings:
    - Add a **Google Analytics Tracking ID** like `G-12ABCDEF3G`
    - Enable **Use Global Site Tag**
    - Enable **Purchase Transactions**
2. Complete an order in the store (NOT AS ADMIN!).

#### Purchase tracking code correctly inserted
4. On the Order Received page (`checkout/order-received/xxx/?key=…`) use the element inspector to confirm that the tracking code is inserted: 
    ```
    gtag( 'event', 'purchase', {
            'transaction_id': '…
    ```
    - Using "View Source" will not work as the page is refreshed and the tracking code no longer inserted.

#### Purchase tracking code correctly not inserted repeatedly
5. Refresh the page, and confirm that the tracking code is no longer visible.
    - WCGA sets the `_ga_tracked` postmeta value for the order to ensure that the transaction is only tracked once.

#### Purchase tracking code correctly inserted if key correct and `_ga_tracked` not set
6. Delete the `_ga_tracked` postmeta value for the order
    ``DELETE FROM `wp_postmeta` WHERE `meta_key`='_ga_tracked' AND post_id = xxx``
7. Refresh the Order Received page and confirm that the tracking code is present again.

#### Purchase tracking code correctly not inserted without key
8. Delete the `key=…` parameter from the query string and refresh the Order Received page.
9. Confirm that the transaction tracking code isn't inserted (in this instance, because of the `_ga_tracked` postmeta value).
10. Delete the `_ga_tracked` postmeta value for the order again
    ``DELETE FROM `wp_postmeta` WHERE `meta_key`='_ga_tracked' AND post_id = xxx``
11. Confirm that the transaction tracking code isn't inserted again (in this instance, because the 
 `key` for the order isn't provided).

### Other information:

### Changelog entry

> Fix - Confirm order key before displaying transaction tracking code.

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -401,6 +401,12 @@ class WC_Google_Analytics extends WC_Integration {
 			return '';
 		}
 
+		// Check order key.
+		$order_key = empty( $_GET['key'] ) ? '' : wc_clean( wp_unslash( $_GET['key'] ) );
+		if ( ! hash_equals( $order->get_order_key(), $order_key ) ) {
+			return '';
+		}
+
 		$load = $this->get_tracking_instance()->load_analytics( $order );
 		$code = $this->get_tracking_instance()->add_transaction( $order );
 
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -403,7 +403,7 @@ class WC_Google_Analytics extends WC_Integration {
 
 		// Check order key.
 		$order_key = empty( $_GET['key'] ) ? '' : wc_clean( wp_unslash( $_GET['key'] ) );
-		if ( ! hash_equals( $order->get_order_key(), $order_key ) ) {
+		if ( ! $order->key_is_valid( $order_key ) ) {
 			return '';
 		}
 
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -402,6 +402,7 @@ class WC_Google_Analytics extends WC_Integration {
 		}
 
 		// Check order key.
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended,WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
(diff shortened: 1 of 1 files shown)


jorgemd24 (approved):
Thanks, @layoutd to fix this security issue!

What I have tested: 

- ✅ I have created a new order and I checked that the tracking code is inserted.
- ✅ I have refreshed the page and I checked that the tracking code is not visible anymore.
- ✅ I have deleted the _ga_tracked postmeta and the tracking code was present again.
- ✅ I have deleted it again, I removed the key parameter and I see that the tracking code is not inserted.
✅ I have tested with older orders, prior to the Google Analytics installation and I can confirm that the tracking code is not inserted if the key is not provided.

includes/class-wc-google-analytics.php
```diff
@@ -401,6 +401,12 @@ protected function get_ecommerce_tracking_code( $order_id ) {
 			return '';
 		}
 
+		// Check order key.
+		$order_key = empty( $_GET['key'] ) ? '' : wc_clean( wp_unslash( $_GET['key'] ) );
```
mikkamp, line 406: Do we need to ignore the phpcs rule `WordPress.Security.NonceVerification.Recommended`
In this case the key would be the nonce.

mikkamp (approved):
Took me a while to get round to the testing, but I can confirm it's all working as well. Nice usage of the helper function since WC core doesn't use that one much.
It's just showing some yellow warnings for phpcs.

layoutd:
Thanks for taking a look, @mikkamp and @jorgemd24! I made a few minor changes for PHPCS happiness (just to the `get_ecommerce_tracking_code` method), and will merge after another quick smoke test.

PR #209, opened 2022-01-06
Fix/activation error when WooCommerce is missing

jorgemd24:
### All Submissions:

In this PR I am proposing to fix the fatal error that was caused when the following situations were happening:

-  the method WC_Google_Analytics_Integration::maybe_show_ga_pro_notices was called.
-  WC was not installed. 
-  the option `woocommerce_google_analytics_pro_notice_shown` was false.

In addition, this PR is showing a warning if WooCommerce is not installed.

Closes # .

1. Add the following line `update_option( 'woocommerce_google_analytics_pro_notice_shown', false );` on top of woocommerce-google-analytics-integration.php
2. Disabled WC plugin.
3. Disabled WC Google Analytics plugin.
4. Enable WC Google Analytics plugin.
5. Not fatal error should be shown and you should see a warning saying "WooCommerce Google Analytics depends on the last version of WooCommerce to work!" .

### Changelog entry

> Fix - Activation error when WC was disabled.

### DIFF
--- a/woocommerce-google-analytics-integration.php
+++ b/woocommerce-google-analytics-integration.php
@@ -26,7 +26,9 @@ if ( ! class_exists( 'WC_Google_Analytics_Integration' ) ) {
 	register_activation_hook(
 		__FILE__,
 		function () {
-			WC_Google_Analytics_Integration::get_instance()->maybe_show_ga_pro_notices();
+			if ( class_exists( 'WooCommerce' ) ) {
+				WC_Google_Analytics_Integration::get_instance()->maybe_show_ga_pro_notices();
+			}
 		}
 	);
 
@@ -43,6 +45,7 @@ if ( ! class_exists( 'WC_Google_Analytics_Integration' ) ) {
 		 */
 		public function __construct() {
 			if ( ! class_exists( 'WooCommerce' ) ) {
+				add_action( 'admin_notices', [ $this, 'woocommerce_missing_notice' ] );
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


ianlin (approved):
Tested locally and worked as expected. 👍 

✅  Use `develop` branch, I can see the fatal error when activating this extension when WC isn’t activated
✅  Use this branch, I can see the admin notice instead of the fatal error when activating this extension when WC isn’t activated

woocommerce-google-analytics-integration.php
```diff
@@ -137,7 +138,7 @@ public function add_integration( $integrations ) {
 		 */
 		public function maybe_show_ga_pro_notices() {
 			// Notice was already shown
-			if ( get_option( 'woocommerce_google_analytics_pro_notice_shown', false ) ) {
+			if ( ! class_exists( 'WooCommerce' ) || get_option( 'woocommerce_google_analytics_pro_notice_shown', false ) ) {
```
budzanowski, line 141: In the `get_option`, the `$default` equals `false` by default so you can omit it.

budzanowski:
Left a very small comment but otherwise looks good. :shipit:

woocommerce-google-analytics-integration.php
budzanowski, line 141: In the `get_option`, the `$default` equals `false` by default so you can omit it.

budzanowski:
Something went wrong with my GH comment, first, it did not comment at all and now I see duplicates. Well, everything is OK.

woocommerce-google-analytics-integration.php
jorgemd24: Thanks, @budzanowski for pointing that out! Would you like me to create a small PR removing this param or how do you proceed in these cases when the branch is already merged?

budzanowski: I would just leave it :) It is something super minor. If you still want you can create a PR with the fix and just merge it without additional review and pointing to this comment.

PR #207, opened 2022-01-03
Fix `composer archive` compatibility issue of composer v2.2+

eason9487:
### All Submissions:

* [x] Does your code follow the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards?
* [ ] (N/A) Have you written new tests for your changes, as applicable?
* [ ] (N/A) Have you successfully run tests with your changes locally?

### Changes proposed in this Pull Request:

Since [composer version 2.2.0](https://github.com/composer/composer/releases/tag/2.2.0), the `.gitignore` file is no longer to be taken into account as ignore patterns when running `composer archive`, this PR adds the missing pattern to ignored list in **.gitattributes** to fix the compatibility issue.

- `node_modules`
- `languages`

### How to test the changes in this Pull Request:

1. Follow the instructions in README to build this extension locally with composer version below **2.2.0**.
2. `composer self-update` to update to the latest version.
3. Build this extension locally again with composer version **2.2+**.
4. Compare the contents of two zip files to see if they are the same.
5. `composer self-update --rollback` if want to roll back to your former version of composer.

### Other information:

* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

PR #213, opened 2022-01-28
Include languages folder and pot file in the release

eason9487:

Fix the problem introduced by #207

The `languages` folder should be always included in the release zip file.

Since `composer` versions 1.x, 2.1.x and 2.2.x have different processing of whether or not to use **.gitignore** as one of ignore patterns source when packaging. This PR makes all these versions include the `languages` folder in the release zip file.

Simple way to update composer to a specific version is to use `composer self-update <version>`

1. `npm ci`
1. `composer self-update 1.10.25`, and `npm run build`
1. Check the created zip file and confirm it contains the languages folder and the .pot file
1. `composer self-update 2.1.9`, and `npm run build`
1. Check the created zip file and confirm it contains the languages folder and the .pot file
1. `composer self-update 2.2.5`, and `npm run build`
1. Check the created zip file and confirm it contains the languages folder and the .pot file

### Changelog entry

> Fix - Include languages folder and POT file in the release.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -24,4 +24,3 @@ docs export-ignore
 wordpress_org_assets export-ignore
 assets/css/*.scss export-ignore
 node_modules export-ignore
-languages export-ignore
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,8 @@
     "license": "GPL-2.0+",
     "archive": {
         "exclude": [
-            "!/assets"
+            "!/assets",
+            "!/languages"
         ]
     }
 }


mikkamp (commented):
WordPress.org extensions don't need to submit the .POT file as it's automatically handled by the translations server.

We can see [here](https://translate.wordpress.org/projects/wp-plugins/woocommerce-google-analytics-integration/stable/en-gb/default/?filters%5Bstatus%5D=untranslated) that the untranslated strings are all from the latest release.

image

Even though the languages folder was not included in the [latest release 1.5.8](https://plugins.svn.wordpress.org/woocommerce-google-analytics-integration/tags/1.5.8/).

Building the translation file as a build task can still be relevant since it will help flag any errors with the domain etc.

So as it is the languages folder can remain excluded for this extension and the PR can just be closed.

eason9487:
@mikkamp, thanks for the information and explanations!

Closing this PR due to WordPress.org extensions don't need to include the .POT file.

Issue #215, opened 2022-04-27
Universal Analytics will stop processing data on July 1, 2023


layoutd:
Google Analytics will no longer collect Universal Analytics data starting on July 1, 2023. As per <https://support.google.com/analytics/answer/11583528>:
> ### Universal Analytics will be going away
> _Last updated: March 16, 2022. Visit this page in the coming months for the latest information._
> - Until July 1, 2023, you can continue to use and collect new data in your Universal Analytics properties.
> - After July 1, 2023, you'll be able to access your previously processed data in your Universal Analytics property for at least six months. We know your data is important to you, and we strongly encourage you to [export](https://support.google.com/analytics/answer/11583528#export) your historical reports during this time.
> - In the coming months, we'll provide a future date for when existing Universal Analytics properties will no longer be available. After this future date, you'll no longer be able to see your Universal Analytics reports in the Analytics interface or access your Universal Analytics data via the API.

**Short Term**
It may make sense to include a notice for users that are currently using a Universal Analytics property in the coming months, so they have time to migrate if desired.

**Long Term**
Refactor/disable/remove functionality associated with Universal Analytics.

layoutd:
Postponing this until changes are about 12 months out, with the idea to add dismissible alerts (for affected users), and then permanent alerts immediately before the changes become effective.

strarsis:
@layoutd: But WooCommerce Google Analytics would work/continue working with the new Google Analytics 4, right?

layoutd:
Hi @strarsis, yes, GA4 will continue to be supported, but users still on Universal Analytics will be prompted to upgrade to GA4 (manually) or lose functionality.

PR #222, opened 2022-08-01
Custom Order Table Compatibility

### LINKED ISSUE #221: Custom Order Table compatibility
### Describe the issue:
To ensure we remain compatible with custom order tables, we should not access order data through post functions, instead we should use the WC_Order functions.

The order meta field `_ga_tracked` uses `get_post_meta` and `update_post_meta`. This should be switched to `WC_Order->get_meta` and `WC_Order->update_meta_data`.

There are also various places in the code where compatibility code is in place for accessing direct order properties. The minimum WooCommerce version has [already been bumped to 3.2](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/69e706ce5da6ada920453a6159480b3ad4054f08). So we can remove any compatibility code for WC < 3.0 when we access any order data.

mikkamp:

This PR add compatibility for the custom order tables by ensuring we use the CRUD functions when handling orders.

Closes #221 

1. Setup tracking with a GA4 account. 
2. Purchase a product and complete the order (as a non admin user).
3. On the order received page confirm that we see a request to https://region1.google-analytics.com/g/collect with `en=purchase`
4. Confirm that the source of the order received page contains a snippet like the following:
5. Refresh the order received page and confirm the purchase event is not sent a second time, postmeta `_ga_tracked` should be set to 1.
```
gtag( 'event', 'purchase', {
	'transaction_id': '98',
	'affiliation': 'domain.test',
	'value': '170.00',
	'tax': '0',
	'shipping': '10',
	'currency': 'EUR',  // Currency,
	'items': [{'id': '19','name': 'Product Name','category': 'Main','price': '80','quantity': '2'},],
} );
```
### Additional details
WC < 3.0 compatibility code was only removed for the GA4 (GTag) implementation. There is already an issue for handling code for Universal Analytics, #215 

### Changelog entry
- Fix - Custom Order table compatibility.

### DIFF
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -164,14 +164,14 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 		}
 		$items .= "]";
 
-		$code  = "" . self::tracker_var() . "( 'event', 'purchase', {
+		$code = self::tracker_var() . "( 'event', 'purchase', {
 			'transaction_id': '" . esc_js( $order->get_order_number() ) . "',
 			'affiliation': '" . esc_js( get_bloginfo( 'name' ) ) . "',
 			'value': '" . esc_js( $order->get_total() ) . "',
 			'tax': '" . esc_js( $order->get_total_tax() ) . "',
 			'shipping': '" . esc_js( $order->get_total_shipping() ) . "',
-			'currency': '" . esc_js( version_compare( WC_VERSION, '3.0', '<' ) ? $order->get_order_currency() : $order->get_currency() ) . "',  // Currency,
-			'items': " . $items . ",
+			'currency': '" . esc_js( $order->get_currency() ) . "',
+			'items': {$items},
 		} );";
 
 		return $code;
@@ -184,7 +184,7 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 	 * @param WC_Order_Item $item    The item to add to a transaction/order
 	 */
 	protected function add_item( $order, $item ) {
-		$_product = version_compare( WC_VERSION, '3.0', '<' ) ? $order->get_product_from_item( $item ) : $item->get_product();
+		$_product = $item->get_product();
 		$variant  = self::product_get_variant_line( $_product );
 
 		$code = "{";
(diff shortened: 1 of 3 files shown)


includes/class-wc-google-gtag-js.php
```diff
@@ -164,14 +164,14 @@ protected function add_transaction_enhanced( $order ) {
...
+		$code = self::tracker_var() . "( 'event', 'purchase', {
 			'transaction_id': '" . esc_js( $order->get_order_number() ) . "',
 			'affiliation': '" . esc_js( get_bloginfo( 'name' ) ) . "',
 			'value': '" . esc_js( $order->get_total() ) . "',
 			'tax': '" . esc_js( $order->get_total_tax() ) . "',
 			'shipping': '" . esc_js( $order->get_total_shipping() ) . "',
-			'currency': '" . esc_js( version_compare( WC_VERSION, '3.0', '<' ) ? $order->get_order_currency() : $order->get_currency() ) . "',  // Currency,
-			'items': " . $items . ",
+			'currency': '" . esc_js( $order->get_currency() ) . "',
+			'items': {$items},
```
puntope, line 174: ❓ Just curious, why we don't escape $items output with `esc_js` as well?

puntope:
❓ 

One thing I'm not sure:

You mentioned that collecting  URL should be: https://region1.google-analytics.com/g/collect 

However I got the request in this URL: https://www.google-analytics.com/g/collect

en=purchase is there and it doesn't send again on refresh

Seems like this is not really important tho... so approved in advance

puntope:
Snippet shows correctly formatted, as well as the `_ga_tracked` is set as 1

puntope (approved):

includes/class-wc-google-gtag-js.php
mikkamp: We should only escape the data within the single quotes, not escape the full `$items` variable. If we look at the `$items` variable it comes from the function `add_item` which uses `esc_js` for each field individually, so no need to run it through `esc_js` a second time.

mikkamp:
> You mentioned that collecting URL should be: https://region1.google-analytics.com/g/collect
> However I got the request in this URL: https://www.google-analytics.com/g/collect

Yes, it's pretty much the same URL, it's just Google using a different URL because I'm in a different region.

Issue #186, opened 2021-03-11
Analytics Scripts should use the WordPress Scripts API


mikejolley:
Scripts added by this plugin are hardcoded in `wp_head`. This makes detection and reuse more complex than it should be. We should instead register and enqueue scripts.

I was going to contribute this, however, it would have caused some other issues due to the way scripts are added/filtered, so instead I will share what I did for someone to implement in the future.

cc @layoutd this would make it easier for the listings plugin to re-use the tag manager script, instead of using filters and string replacements.

This needs work but the basic idea is to register the script with a handle, then use inline scripts to initialise with configs.

**Google Tag Manager:**

```php
wp_register_script( 'google-tag-manager', 'https://www.googletagmanager.com/gtag/js?id=' . self::get( 'ga_id' ), array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, false );

wp_add_inline_script(
				'google-tag-manager',
				"
				window.dataLayer = window.dataLayer || [];
				function gtag(){dataLayer.push(arguments);}
				gtag('js', new Date());
				gtag('config', '" . esc_js( $ga_settings['ga_id'] ) . "', { 'send_page_view': false });"
);
```

**Classic and Universal:**

```php
if ( 'yes' === self::get( 'ga_use_universal_analytics' ) ) {
			wp_register_script( 'google-analytics', '//google-analytics.com/analytics.js', array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
			wp_add_inline_script(
				'google-analytics',
				'window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;',
			);
} else {
			wp_register_script( 'google-analytics', 'yes' === self::get( 'ga_support_display_advertising' ) ? 'https://stats.g.doubleclick.net/dc.js' : '//google-analytics.com/ga.js', array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
			// Inline script todo
}
```

Then we can make them async:

```
add_filter( 'script_loader_tag', array( $this, 'async_script_loader_tag' ) );
```

```
	/**
	 * Add async to script tags with defined handles.
	 *
	 * @param string $tag HTML for the script tag.
	 * @param string $handle Handle of script.
	 * @param string $src Src of script.
	 * @return string
	 */
	public function async_script_loader_tag( $tag, $handle, $src ) {
[...]

PR #230, opened 2022-09-30
Fix inconsistencies across item data in events

### LINKED ISSUE #200: Inconsistent names for variable products
#### What I expected

Variable product names to be consistent in Google Analytics through the purchase process, i.e. on whether the variation attributes are included in the name.

#### What happened instead

Add to Cart:

`{items: [{id: "#53", name: "Variable Product", quantity: "1"}]}`

Checkout:

`{items: [{id: "#57", name: "Variable Product", category: "Uncategorized", variant: "size: Medium, color: Red", price: "32", quantity: "3"}]}`

Purchase:

`{items: [{id: "57", name: "Variable Product - Medium, Red", category: "Uncategorized", variant: "size: Medium, color: Red", price: "32", quantity: "3"}]}`

Notice how the variant details are included in the name in the Purchase, but not in others? Seems inconsistent, and might cause awkward looking data in GA.

#### Steps to reproduce the issue

I had trouble getting GA set up correctly to repro this, so @layoutd helped me out in Slack, and directed me here.

Originally brought up in this forums thread:

https://wordpress.org/support/topic/attributes-difference-between-addtocart-and-purchase/?view=all

... and discussed in Slack here:

p1633008284044100-slack-C7U3Y3VMY

-------------------

- [ ] Issue assigned to next milestone.
- [ ] Issue assigned a priority (will be assessed by maintainers).

martynmjones:

This PR fixes inconsistencies across how data is structured for different events. Closes #200 

A `get_product_identifier` method is [introduced](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d8ba3cee842fd45b3f854f19b989cc93a4e6813c/includes/class-wc-abstract-google-analytics-js.php#L116) and used for the item ID across [all events](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/0d6fdf7df9145956fde7e2d405cece393dd6b68f). The ID will either be the variation SKU, product SKU, or # + Product ID.

The [category line](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/ece7d75210502ed581967b0299521a29efcc9c88) is included with the add to cart event and the item name will always use the product title instead of the [variation name](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/fe67b9e69c9e6ea976979e880829fd6de16d7cc8) when an order is placed.

Additionally, Javascript is now used to build the [item ID and variant](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/81d30e57fd950dda15cbfa9c1cd656a36b797c76) when a product is added to cart to make sure the variation selection is considered.

1. Ensure [Google Analytics Debugger](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) is installed and ON
2. Create a variable product
3. Add product to cart
4. View developer console and check the item ID, Name, Category, and Variant
5. Go to checkout
6. View developer console and check the item ID, Name, Category, and Variant. These should all match the values from step 4
8. Place order using any payment method
7. View developer console and check the item ID, Name, Category, and Variant. These should all match the values from step 4 and 6


includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -107,6 +107,20 @@ public function add_transaction( $order ) {
 	 */
 	abstract protected function add_transaction_enhanced( $order );
 
+	/**
+	 * Get item identifier from product data
+	 *
+	 * @param  WC_Product $product WC_Product Object
+	 * @return stirng
```
mikkamp: Small typo here

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -107,6 +107,20 @@ public function add_transaction( $order ) {
...
+	 */
+	public static function get_product_identifier( $product ) {
```
mikkamp, line 116: I can understand why it's made static, since that's the pattern that's followed for other "helper" functions in this class. Although you could argue that making it `static` doesn't really provide any benefits and will only make it harder to unit test is we add unit tests for this code.

However, why do we need to make it public? As far as I can see it's only called from a class that's extends the abstract class, so `protected` will work fine here.

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -162,7 +176,7 @@ protected function add_transaction_universal( $order ) {
 	 * @param  WC_Product $_product  Product to pull info for
 	 * @return string                Line of JSON
 	 */
-	protected static function product_get_category_line( $_product ) {
+	public static function product_get_category_line( $_product ) {
```
mikkamp, line 179: I can see why it's made public because it's called from outside the class, but it's called using the instance there, so that doesn't really dictate why this function should remain static.

includes/class-wc-google-analytics-js.php
```diff
@@ -120,7 +120,7 @@ public static function listing_impression( $product, $position ) {
 
 		wc_enqueue_js( "
 			" . self::tracker_var() . "( 'ec:addImpression', {
-				'id': '" . esc_js( $product->get_id() ) . "',
+				'id': '" . self::get_product_identifier( $product ) . "',
```
mikkamp: The changes in this file are for Universal Analytics, which is still around for backward compatibility. I'm a bit on the fence whether it's useful to also make changes to the older implementation. I'd assume it would have less of an impact to only make the changes for sites using the GA4 implementation.

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
 		$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : "#" . $product->get_id() ) . "'";
 
 		if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
+			// Update SKU and attributes when variation selection changes
+			wc_enqueue_js(
```
mikkamp: Since this is just static JS code it would be nicer if we would use [`wp_enqueue_script`](https://developer.wordpress.org/reference/functions/wp_enqueue_script/). For passing dynamic data like the product SKU [`wp_localize_script`](https://developer.wordpress.org/reference/functions/wp_localize_script/) can be used.

The main advantage of this change would be that the browser could load this script from cache so it doesn't need to be injected on each product page (note that it's currently also being included for simple products). It also allows a bit more flexibility for other plugins, which is what we need to work towards as outlined [here](https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/186).

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
...
+					}
+					return variant.substring( 0, variant.length -1 )
+				}'
+			);
+
 			$item = "{";
-			$item .= "'id': '" . esc_js( $product->get_sku() ? $product->get_sku() : ( '#' . $product->get_id() ) ) . "',";
+			$item .= "'id': ( 0 !== wc_ga_integration_variation_sku.length ) ? ( wc_ga_integration_variation_sku ) : ( '#' + $('.woocommerce-variation-add-to-cart input[name=\"product_id\"]').val() ),";
 			$item .= "'name': '" . esc_js( $product->get_title() ) . "',";
+			$item .= "'category': " . $this->get_tracking_instance()->product_get_category_line( $product );
```
mikkamp: I'm not convinced that we should share this code between the two tracking instances. Technically this should be `item_id, item_name, item_category` as shown in the docs: https://developers.google.com/analytics/devguides/collection/ga4/reference/events#add_to_cart

It's the same for other events, so maybe we should report that in a separate issue.

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -107,6 +107,20 @@ public function add_transaction( $order ) {
...
+	 * Get item identifier from product data
+	 *
+	 * @param  WC_Product $product WC_Product Object
+	 * @return stirng
+	 */
+	public static function get_product_identifier( $product ) {
+		if ( ! empty( $product->get_sku() ) ) {
+			return esc_js( $product->get_sku() );
+		} else {
+			return esc_js( '#' . ( $product->is_type( 'variation' ) ? $product->get_parent_id() : $product->get_id() ) );
```
mikkamp, line 120: Previously it would have sent the variation ID when it encountered a variation, such as when it's called on checkout. Do we want to change that behaviour? In most other circumstances it would already be using the parent product.

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
...
+					}
+				});
+
+				function wc_ga_integration_format_variant() {
+					var variant = "";
+					for ( key in wc_ga_integration_variation_attributes ) {
+						variant += key.replace( "attribute_", "" ) + ": " + wc_ga_integration_variation_attributes[key] + ","
+					}
+					return variant.substring( 0, variant.length -1 )
+				}'
```
mikkamp: I'm not a JS pro, but I believe when looping through the object properties it's best to use the hasOwnProperty check like is done here: https://github.com/woocommerce/woocommerce/blob/6.9.4/plugins/woocommerce/client/legacy/js/frontend/add-to-cart-variation.js#L766

I see the prefix `attribute_` is getting removed, but on my site it uses a global product attribute with the prefix `pa_`. So it's tracked as `pa_colour: blue`

Also what do you think about making `variant` an array. Which means we can just call:

```js
return variants.join();
```
To me that looks cleaner than adding the string plus comma and later removing the trailing comma. Also if `wc_ga_integration_variation_attributes` would somehow be an empty object then there wouldn't be an extra trailing comma.

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
...
+			);
+
 			$item = "{";
-			$item .= "'id': '" . esc_js( $product->get_sku() ? $product->get_sku() : ( '#' . $product->get_id() ) ) . "',";
+			$item .= "'id': ( 0 !== wc_ga_integration_variation_sku.length ) ? ( wc_ga_integration_variation_sku ) : ( '#' + $('.woocommerce-variation-add-to-cart input[name=\"product_id\"]').val() ),";
```
mikkamp: This check fails when we try this for a simple product, because we can't check the length of a variable that hasn't been defined yet. It would be better to initialise the variable as an empty string.

mikkamp (commented):
Thanks for the changes, the code is working as expected.

I've gone through the code and added some suggestions about different things I noticed. Feel free to move any of those to a follow up issue if it's not within the scope of this PR.

includes/class-wc-google-analytics.php
tomalec: +1 to the above.
Having it as `.js` file could also benefit in better IDE support, linting, testability etc.

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
 		$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : "#" . $product->get_id() ) . "'";
 
 		if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
+			// Update SKU and attributes when variation selection changes
+			wc_enqueue_js(
+				'var wc_ga_integration_product_sku,
+					wc_ga_integration_variation_sku,
+					wc_ga_integration_variation_attributes;
+
+				wc_ga_integration_product_sku = "' . esc_js( $product->get_sku() ) . '";
```
tomalec: :construction: 

```suggestion
				'const wc_ga_integration_product_sku = "' . esc_js( $product->get_sku() ) . '";
				let wc_ga_integration_variation_sku, wc_ga_integration_variation_attributes;				
```

Also, are the below functions ok, with those being `undefined`?

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
...
+
+				$(document).on( "found_variation", "form.cart", function( e, variation ) {
+					wc_ga_integration_variation_attributes = variation.attributes;
+					if ( "undefined" !== typeof variation.sku && 0 !== variation.sku.length ) {
```
tomalec: :nail_care: 
Speaking of `.js` vs `.php` linting. In JS we avoid yoda conditions

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
...
+						wc_ga_integration_variation_sku = variation.sku;
+					} else {
+						wc_ga_integration_variation_sku = wc_ga_integration_product_sku;
+					}
+				});
+
+				function wc_ga_integration_format_variant() {
+					var variant = "";
+					for ( key in wc_ga_integration_variation_attributes ) {
```
tomalec: :construction:
```suggestion
					for ( const key in wc_ga_integration_variation_attributes ) {
```

includes/class-wc-google-analytics.php
```diff
@@ -482,10 +482,42 @@ public function add_to_cart() {
...
+						variant += key.replace( "attribute_", "" ) + ": " + wc_ga_integration_variation_attributes[key] + ","
+					}
+					return variant.substring( 0, variant.length -1 )
```
tomalec: Personally, I'd use an array to push entries to , then use `.join(',')`. I find it easier to get the high-level idea of what's happening that string manipulations.

Also, AFAIR `.join` micro performs better than string append (`+=`).

tomalec: BTW, there is a semicolon missing.

tomalec:
Could you explain what, when, and why this code does?
I'm afraid I don't fully get the lifecycle here, and why we construct JS code in PHP. Also why we build it and serve in two separate parts. One with static function https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/fix/200-inconsistent-item-data/includes/class-wc-google-analytics.php#L486 and the other with JSON like object that executes some functions https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/fix/200-inconsistent-item-data/includes/class-wc-google-analytics.php#L518. I think what bothers me here, even more, is when the latter is executed?



Without knowing that, I'm afraid we can get into some timing/racy issues as `wc_ga_integration_format_variant` function relies on some data being populated asynchronously after some event is fired.

assets/js/ga-integration.js
```diff
@@ -0,0 +1,5 @@
+jQuery(document).ready( function($) {
+    $(document).on( "found_variation", "form.cart", function( e, variation ) {
+        google_analytics_integration = variation.google_analytics_integration;
```
tomalec: It's a good practice to avoid polluting the global scope. Such a variable could me accidentally, or maliciously changed by some other entity.
I'd rather limit its availability, either by attaching it to some element, object, or sharing a JS scope with

includes/class-wc-abstract-google-analytics-js.php
martynmjones: You're right that `protected` would have worked fine here, however, in the last update it is now being called from [class-wc-google-analytics.php](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2754e1955f0779f9e090eaa6df1cc4ea3cb8635d/includes/class-wc-google-analytics.php#L492) so it will now need to remain public.

includes/class-wc-google-analytics-js.php
martynmjones: That makes perfect sense! I've [rolled back](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/230/commits/cf5ee3c1ef844a3af4ee6dbff6c937fef1dd8d5f) the changes made to this file.

includes/class-wc-google-analytics.php
martynmjones: Thank you both for the input here. I've reworked the approach and the JS is far simpler now: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/230/commits/1ee7139c8e1e0e7b997491bb8ec746898c73f54b

includes/class-wc-abstract-google-analytics-js.php
martynmjones: Currently, the variation ID is sent on the checkout started event but it isn't used for the add to cart event. The data structure will have to change for some events so that we have consistency across all tracking.

Given that we're also specifying the `variant` when sending I don't think it makes sense to use the variation ID as the item ID which is why I left it out here.

includes/class-wc-google-analytics.php
martynmjones: As the formatted strings are now [coming from the backend](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/cf5ee3c1ef844a3af4ee6dbff6c937fef1dd8d5f/includes/class-wc-google-analytics.php#L488) there isn't an issue between simple and variable products anymore.

includes/class-wc-google-analytics.php
martynmjones: Good to know! Thanks 👍

includes/class-wc-google-analytics.php
martynmjones: I've opened a separate issues so this can be investigated: https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/233

includes/class-wc-abstract-google-analytics-js.php
mikkamp: In my opinion the `add_to_cart` event code is badly placed in the main class. As far as I can see it's outputting a JS snippet just like the rest of the events. But that's outside the scope of this PR, so fine to leave as is.

includes/class-wc-google-gtag-js.php
```diff
@@ -28,6 +31,33 @@ public static function get_instance( $options = array() ) {
...
+	 */
+	public function setup_frontend_scripts() {
+		global $product;
+
+		if ( $product->is_type( 'variable' ) ) {
+			// Filter variation data to include formatted strings required for add_to_cart event
+			add_filter( 'woocommerce_available_variation', array( $this, 'variant_data' ), 10, 3 );
+			// Add default inline product data for add to cart tracking
+			wp_enqueue_script( $this->script_handle );
+			wp_add_inline_script( $this->script_handle, $this->default_add_to_cart_data( $product ) );
```
mikkamp: Shouldn't we be passing `before` as the third argument? Not sure if you are going to refactor, but the scripts are sharing a global variable, so we need to make sure we define it first.

includes/class-wc-google-gtag-js.php
```diff
@@ -28,6 +31,33 @@ public static function get_instance( $options = array() ) {
...
+		}
+	}
+
+	/**
+	 * Register front end JavaScript
+	 */
+	public function register_scripts() {
+		wp_register_script( $this->script_handle, plugins_url( 'assets/js/ga-integration.js', dirname( __FILE__ ) ), array( 'jquery' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
```
mikkamp, line 59: Because of the way `plugin_basename` works internally this still works but what we are doing is passing the path `...wp-content/plugins/woocommerce-google-analytics-integration/includes`. In this case I think it's better to pass `dirname( __DIR__ )` which will pass `...wp-content/plugins/woocommerce-google-analytics-integration`

mikkamp (commented):
Thanks for the additional changes. It's looking good and working as intended, I just left a few smaller comments.

In regards to the comment from @tomalec about not polluting the global scope. If we would have used `wp_localize_script` it would have added the data to a global variable, which traditionally was a very common practice. For this PR we can modify the JS code in the `add_to_cart` function to retrieve the data from the element it's attached to and otherwise fallback to a default based on the product data. That will eliminate the need for calling `wp_add_inline_script` to set the defaults in a global.

Eventually we want to work towards a generic JS function which tracks an event, which we can call directly from the `found_variation` hook and passes the variation data to the event function, so we don't need to rely on the inline script, but that's beyond the scope of this PR.

assets/js/ga-integration.js
martynmjones: As we discussed, it's now storing the [data in an array](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/230/commits/e424a53bcf7e4dc60091db7ef0b46777c4fbb49f) and then retrieves the correct data based on the currently selected variation when a product is added to a cart. It's not the most elegant solution but as @mikkamp said, when the extension moves toward a generic JS function for event tracking we'll be able to avoid globals.

includes/class-wc-google-gtag-js.php
martynmjones: In the latest update, I've [removed the inline script for product defaults](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/230/commits/4e4f24f5791437262ab320f94ffd67d544665220) here so this is no longer an issue.

includes/class-wc-google-gtag-js.php
martynmjones: I wasn't aware of the differences in `plugin_basename` so TIL and thanks for pointing that out, however, using `plugins_url` here `dirname( __DIR__ )` would result in a broken path for the script so it would need to use `__DIR__` or `dirname( __FILE__ )`

mikkamp: Ok, I see what's happening there, `plugins_url` is asking for any file within the root plugin directory. It strips off the file to get the directory name. So `includes` is seen as the file which is stripped. A bit confusing in my opinion but it works so we can leave it as is. In other plugins the main plugin file is passed to a class through a constant: https://github.com/woocommerce/automatewoo/blob/5.6.0/automatewoo.php#L41

mikkamp (approved):
Thanks for the changes, all the testing I did is working as intended. I left one comment about the js loading path, but fine to leave as is. I don't see anything else obvious so I think this is good to go.

PR #235, opened 2022-10-31
Add Google Analytics into Things to do next list

jorgemd24:

Closes  peeuvX-6w-p2

This PR adds a new task in WooCommerce's "Things to do next" section.

image

The configuration is completed once is set the "Google Analytics Tracking ID".

image

In order to give a more consistent experience through the different extensions the following banner has been removed:

image

### Checks:

* [X] Does your code follow the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards?
* [X] Have you successfully run tests with your changes locally?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

1. Go to Woocomme Dashboard - wp-admin/admin.php?page=wc-admin
2. A new task should be displayed: Set up Google Analytics
3. Google Analytics should be completed if a "Google Analytics Tracking ID" has been set previously otherwise will be uncompleted.
4. Click on the task. It should go to the settings page.
5. Add or remove "Google Analytics Tracking ID" and the status of the task should change to complete/uncompleted.
6. No banners should be displayed if the Google Analytics Tracking ID is not set.

### Aditional details

I thought to delete `class-wc-google-analytics-info-banner.php` as is not used anymore in the plugin however I am not sure if other plugins are using this class so I decided to keep the file to be safe.

### DIFF
--- a/includes/class-wc-google-analytics-task.php
+++ b/includes/class-wc-google-analytics-task.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * Set up Google Analytics Integration task.
+ *
+ * Adds a set up Google Analytics Integration task to the task list.
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+use Automattic\WooCommerce\Admin\Features\OnboardingTasks\Task;
+
+/**
+ * Setup Task class.
+ *
+ * @phpcs:disable Squiz.Classes.ClassFileName.NoMatch
+ */
+class GoogleAnalyticsTask extends Task {
+
+	/**
+	 * Get the ID of the task.
+	 *
+	 * @return string
+	 */
+	public function get_id() {
+		return 'setup-google-analytics-integration';
+	}
+
+	/**
+	 * Get the title for the task.
+	 *
+	 * @return string
+	 */
+	public function get_title() {
+		return esc_html__( 'Set up Google Analytics', 'woocommerce-google-analytics-integration' );
+	}
+
+	/**
+	 * Get the content for the task.
+	 *
+	 * @return string
+	 */
+	public function get_content() {
+		return esc_html__( 'Provides the integration between WooCommerce and Google Analytics.', 'woocommerce-google-analytics-integration' );
+	}
+
+	/**
+	 * Get the time required to perform the task.
+	 *
+	 * @return string
+	 */
+	public function get_time() {
+		return esc_html__( '5 minutes', 'woocommerce-google-analytics-integration' );
+	}
+
+	/**
+	 * Get the action URL for the task.
+	 *
+	 * @return string
+	 */
+	public function get_action_url() {
+		return WC_Google_Analytics_Integration::get_instance()->get_settings_url();
+	}
+
+	/**
+	 * Check if the task is complete.
+	 *
+	 * @return bool
+	 */
+	public function is_complete() {
+		return WC_Google_Analytics_Integration::get_integration()->is_setup_complete();
+	}
+
+}
+
+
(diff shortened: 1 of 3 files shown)


includes/class-wc-google-analytics-task.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/**
+ * Set up Google Analytics Integration task.
+ *
+ * Adds a set up Google Analytics Integration task to the task list.
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+use Automattic\WooCommerce\Admin\Features\OnboardingTasks\Task;
```
martynmjones, line 10: On a site with <= WC version 5.8 installed this triggers a critical error because `OnboardingTasks\Task` wasn't available at that point but the extension is currently checking for a [minimum version of WC 3.2](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/617ec963d512ce454d1e665fe76cf89530471b74/woocommerce-google-analytics-integration.php#L57).

It also looks like there is an issue with abstract method declarations from WC version 6.1.0 which also triggers a fatal error.

martynmjones:
Looks like this is working great with WooCommerce version 6.4.0 onward 👍. There are a couple of backward compatibility issues that I mentioned in a comment.

jorgemd24:
Thanks, @martynmjones for having a look at this, 

>There are a couple of backward compatibility issues that I mentioned in a comment.

Can you point me to the comment? I can not see it 😄 

Thanks!

martynmjones (changes requested):

martynmjones:
> Thanks, @martynmjones for having a look at this,
> 
> > There are a couple of backward compatibility issues that I mentioned in a comment.
> 
> Can you point me to the comment? I can not see it 😄
> 
> Thanks!

Sorry Jorge, didn't submit the [review](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/235/files/617ec963d512ce454d1e665fe76cf89530471b74#r1010302740) 🥱 ☕

jorgemd24:
@martynmjones thanks for catching this.

>On a site with <= WC version 5.8 installed this triggers a critical error because OnboardingTasks\Task wasn't available at that point but the extension is currently checking for a [minimum version of WC 3.2](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/617ec963d512ce454d1e665fe76cf89530471b74/woocommerce-google-analytics-integration.php#L57).

I did some investigation and I can see that the issue is because WC 5.8.0 uses   [WC admin. V2.7.2](https://github.com/woocommerce/woocommerce/blob/d9620e89de7d473ac2f8962091dbe6d4990d76cf/composer.json#L24) and the class was introduced in V2.8.0

>It also looks like there is an issue with abstract method declarations from WC version 6.1.0 which also triggers a fatal error.

WC 6.1.0 uses V3.0.3 and it has an abstract method [get_parent_id](https://github.com/woocommerce/woocommerce-admin/blob/0b1abab35c62717f3121428fe7f378998ef4899c/src/Features/OnboardingTasks/Task.php#L75) than in the later versions has been implemented in the class, so the method is not abstract anymore.

I think it is a good opportunity to update the minimum WC version required to run the plugin, as I think we need to be aligned with the [L2 WC](https://developer.woocommerce.com/2020/05/08/wordpress-core-support-policy/) and [the team's policy](https://automatap2.wordpress.com/how-we-work/environment-developer-distribution/) meaning that we support the current version and the last two versions of WC.

I have updated the code with the WC minimum version and the error message :

image

martynmjones (approved):

martynmjones:
> I think it is a good opportunity to update the minimum WC version required to run the plugin, as I think we need to be aligned with the [L2 WC](https://developer.woocommerce.com/2020/05/08/wordpress-core-support-policy/) and [the team's policy](PeeuvX-1t-p2) meaning that we support the current version and the last two versions of WC.
> 
> I have updated the code with the WC minimum version and the error message :
> 
> image

Thanks for the detailed response! No longer seeing any errors triggered by the extension regardless of WC version so LGTM 👍

Issue #185, opened 2021-03-11
Inconsistent usage of woocommerce_ga_tracker_variable


mikejolley:
A filter named `woocommerce_ga_tracker_variable` exists which allows `gtag` and `ga` to be changed. Some calls to Google Analytics then use this, e.g:

```
' . self::tracker_var() . "( 'ec:addImpression', {
```

However, usage is inconsistent. There are many hardcoded gtag and ga calls through the codebase. If the filter is used, these calls can break.

I'd suggest removing the filter and using the correct methods throughout.

PR #240, opened 2022-11-29
Add common events method

martynmjones:

This PR adds a [common method](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/cdc92f526fe095f3c31cf9c4b63cc56aa624d4cb/includes/class-wc-google-gtag-js.php#L95) that is used to generate gtag javascript event code and format data for that event.

The main benefits this brings to the extension is:

1. Consistent usage of `woocommerce_gtag_tracker_variable` filter #185 
2. Adds a [filter](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/cdc92f526fe095f3c31cf9c4b63cc56aa624d4cb/includes/class-wc-google-gtag-js.php#L106) for event data which can be used in unit tests and by third-party developers to manipulate event data #171 

### Checks:

- [x] Does your code follow the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards?
- [ ] Have you written new tests for your changes, as applicable?
- [x] Have you successfully run tests with your changes locally?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/pulls) for the same update/change?

1. Setup extension with `Enable Enhanced eCommerce` enabled
2. Checkout branch
3. Make sure [Google Analytics Debugger](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) is **on**
4. In the developer console make sure `Preserve log` is enabled
5. View main product archive with test products published
6. Filter console for `view_item_list` and check event config and data matches current structure (_See below for table detailing existing event data_)
7. Click on a product
8. Filter console for `select_content` and check event config and data matches current structure
9. Filter console for `view_item` and check event config and data matches current structure
10. Add product to cart and then goto checkout page
11. Filter console for `begin_checkout` and check event config and data matches current structure
12. Place order
13. Filter console for `purchase` and check event config and data matches current structure
14. Visit product archive, add a product to cart, and then remove the item from cart
15. Filter console for `remove_from_cart` and check event config and data matches current structure

Markup 2022-12-01 at 22 08 23

### Additional Details

An [existing method](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/cdc92f526fe095f3c31cf9c4b63cc56aa624d4cb/includes/class-wc-google-gtag-js.php#L406) has similar functionality, however, it insists on an element selector and javascript click event to trigger the Google Analytics event. While there is some aspect of duplication in this PR I have left the existing function as is because we will eventually want to move the `add_to_cart` functionality out of the abstract class and have it follow the same structure as the events in this PR.

Additionally, a follow-up PR will be required to address other [inconsistent usage of the `woocommerce_gtag_tracker_variable` filter](https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/185).

<hr />
<details>
<summary><b>View table of existing gtag events</b></summary>
<br />
<p><i>This PR doesn't aim to change any of the existing data structures in the extension and so they should exactly match what currently exists. This table outlines what data is currently being sent with each event in <a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/tree/d855d7ef7e80b1623efa7873740a3d6abdeab949">trunk</a>.</i></p>
<table>
<thead>
<tr>
<th>Event</th>
<th>Existing data</th>
</tr>
</thead>
<tbody>
<tr>
<td><a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d855d7ef7e80b1623efa7873740a3d6abdeab949/includes/class-wc-google-gtag-js.php#L48">view_item_list</a></td>
<td>
<pre><code>gtag( 'event', 'view_item_list', { 'items': [ {
  'id': '',
  'name': '',
  'category': '',
  'list': '',
  'list_position': ''
} ] } );</code></pre>
</td>
</tr>
<tr>
<td><a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d855d7ef7e80b1623efa7873740a3d6abdeab949/includes/class-wc-google-gtag-js.php#L72">select_content</a></td>
<td>
<pre><code>gtag( 'event', 'select_content', {
  'content_type': 'product',
  'items': [ {
    'id': '',
    'name': '',
    'category': '',
    'list_position': ''
  } ],
} );</code></pre>
</td>
</tr>
<tr>
<td><a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d855d7ef7e80b1623efa7873740a3d6abdeab949/includes/class-wc-google-gtag-js.php#L231">view_item</a></td>
<td>
<pre><code>gtag( 'event', 'view_item', {
  'content_type': 'product',
  'items': [ {
    'id': '',
    'name': '',
    'category': '',
    'price': ''
  } ],
} );</code></pre>
</td>
</tr>
<tr>
<td><a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d855d7ef7e80b1623efa7873740a3d6abdeab949/includes/class-wc-google-gtag-js.php#L252">begin_checkout</a></td>
<td>
<pre><code>gtag( 'event', 'begin_checkout', {
  'content_type': 'product',
  'items': [ {
    'id': '',
    'name': '',
    'category': '',
    'variant': '', (conditional)
    'price': '',
    'quantity': '',
  } ],
} );</code></pre>
</td>
</tr>
<tr>
<td><a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d855d7ef7e80b1623efa7873740a3d6abdeab949/includes/class-wc-google-gtag-js.php#L157">purchase</a></td>
<td>
<pre><code>gtag( 'event', 'purchase', {
  'transaction_id': '',
  'affiliation': '',
  'value': '',
  'tax': '',
  'shipping': '',
  'currency': 'USD',
  'items': [{
    'id': '',
    'name': '',
    'category': '',
    'price': '',
    'quantity': ''
  },],
} );</code></pre>
</td>
</tr>
<tr>
<td><a href="https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d855d7ef7e80b1623efa7873740a3d6abdeab949/includes/class-wc-google-gtag-js.php#L209"> remove_from_cart</a></td>
<td>
<pre><code>gtag( 'event', 'remove_from_cart', {
  'items': [ {
    'id': '',
    'quantity': '',
  } ]
} );</code></pre>
</td>
</tr>
</tbody>
</table>
</details>
<hr />

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -465,7 +465,7 @@ class WC_Google_Analytics extends WC_Integration {
 	/**
 	 * Google Analytics event tracking for single product add to cart
 	 */
-	public function add_to_cart() {
+	public function add_to_cart() { 
 		if ( $this->disable_tracking( $this->ga_event_tracking_enabled ) ) {
 			return;
 		}
@@ -475,35 +475,39 @@ class WC_Google_Analytics extends WC_Integration {
 
 		global $product;
 
-		// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
-		$parameters = array();
-		$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
-		$parameters['action']   = "'" . __( 'Add to Cart', 'woocommerce-google-analytics-integration' ) . "'";
-		$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : "#" . $product->get_id() ) . "'";
+		if ( 'yes' === $this->ga_gtag_enabled ) {
+			$this->get_tracking_instance()->add_to_cart( $product );
+		} else {
+			// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
+			$parameters = array();
+			$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
+			$parameters['action']   = "'" . __( 'Add to Cart', 'woocommerce-google-analytics-integration' ) . "'";
+			$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : "#" . $product->get_id() ) . "'";
 
-		if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
+			if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
 
-			$item = "{";
+				$item = "{";
 
-			if ( $product->is_type( 'variable' ) ) {
-				$item .= "'id': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].id : false,";
-				$item .= "'variant': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].variant : false,";
-			} else {
-				$item .= "'id': '" . $this->get_tracking_instance()->get_product_identifier( $product ) . "',";
+				if ( $product->is_type( 'variable' ) ) {
+					$item .= "'id': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].id : false,";
+					$item .= "'variant': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].variant : false,";
+				} else {
+					$item .= "'id': '" . $this->get_tracking_instance()->get_product_identifier( $product ) . "',";
+				}
+
+				$item .= "'name': '" . esc_js( $product->get_title() ) . "',";
+				$item .= "'category': " . $this->get_tracking_instance()->product_get_category_line( $product );
+				$item .= "'quantity': $( 'input.qty' ).val() ? $( 'input.qty' ).val() : '1'";
+				$item .= "}";
+
+				$parameters['item'] = $item;
+
+				$code = "" . $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . " );";
+				$parameters['enhanced'] = $code;
 			}
 
-			$item .= "'name': '" . esc_js( $product->get_title() ) . "',";
-			$item .= "'category': " . $this->get_tracking_instance()->product_get_category_line( $product );
-			$item .= "'quantity': $( 'input.qty' ).val() ? $( 'input.qty' ).val() : '1'";
-			$item .= "}";
-
-			$parameters['item'] = $item;
-
-			$code = "" . $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . " );";
-			$parameters['enhanced'] = $code;
+			$this->get_tracking_instance()->event_tracking_code( $parameters, '.single_add_to_cart_button' );
 		}
-
-		$this->get_tracking_instance()->event_tracking_code( $parameters, '.single_add_to_cart_button' );
 	}
 
 	/**
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -479,7 +479,7 @@ class WC_Google_Analytics extends WC_Integration {
 			$this->get_tracking_instance()->add_to_cart( $product );
 			return;
 		}
-		
+
 		// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
 		$parameters             = array();
 		$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
@@ -504,7 +504,7 @@ class WC_Google_Analytics extends WC_Integration {
 
 			$parameters['item'] = $item;
 
-			$code                   = $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . ' );';
+			$code                   = $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', {$item} );";
 			$parameters['enhanced'] = $code;
 		}
 
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -465,7 +465,7 @@ class WC_Google_Analytics extends WC_Integration {
 	/**
 	 * Google Analytics event tracking for single product add to cart
 	 */
-	public function add_to_cart() { 
+	public function add_to_cart() {
 		if ( $this->disable_tracking( $this->ga_event_tracking_enabled ) ) {
 			return;
 		}
@@ -479,14 +479,14 @@ class WC_Google_Analytics extends WC_Integration {
 			$this->get_tracking_instance()->add_to_cart( $product );
 		} else {
 			// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
-			$parameters = array();
+			$parameters             = array();
 			$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
 			$parameters['action']   = "'" . __( 'Add to Cart', 'woocommerce-google-analytics-integration' ) . "'";
-			$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : "#" . $product->get_id() ) . "'";
+			$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : '#' . $product->get_id() ) . "'";
 
 			if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
 
-				$item = "{";
+				$item = '{';
 
 				if ( $product->is_type( 'variable' ) ) {
 					$item .= "'id': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].id : false,";
@@ -498,11 +498,11 @@ class WC_Google_Analytics extends WC_Integration {
 				$item .= "'name': '" . esc_js( $product->get_title() ) . "',";
 				$item .= "'category': " . $this->get_tracking_instance()->product_get_category_line( $product );
 				$item .= "'quantity': $( 'input.qty' ).val() ? $( 'input.qty' ).val() : '1'";
-				$item .= "}";
+				$item .= '}';
 
 				$parameters['item'] = $item;
 
-				$code = "" . $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . " );";
+				$code                   = $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . ' );';
 				$parameters['enhanced'] = $code;
 			}
 
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -477,37 +477,38 @@ class WC_Google_Analytics extends WC_Integration {
 
 		if ( 'yes' === $this->ga_gtag_enabled ) {
 			$this->get_tracking_instance()->add_to_cart( $product );
-		} else {
-			// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
-			$parameters             = array();
-			$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
-			$parameters['action']   = "'" . __( 'Add to Cart', 'woocommerce-google-analytics-integration' ) . "'";
-			$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : '#' . $product->get_id() ) . "'";
+			return;
+		}
+		
+		// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
+		$parameters             = array();
+		$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
+		$parameters['action']   = "'" . __( 'Add to Cart', 'woocommerce-google-analytics-integration' ) . "'";
+		$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : '#' . $product->get_id() ) . "'";
 
-			if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
+		if ( ! $this->disable_tracking( $this->ga_enhanced_ecommerce_tracking_enabled ) ) {
 
-				$item = '{';
+			$item = '{';
 
-				if ( $product->is_type( 'variable' ) ) {
-					$item .= "'id': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].id : false,";
-					$item .= "'variant': google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ? google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].variant : false,";
-				} else {
-					$item .= "'id': '" . $this->get_tracking_instance()->get_product_identifier( $product ) . "',";
-				}
-
-				$item .= "'name': '" . esc_js( $product->get_title() ) . "',";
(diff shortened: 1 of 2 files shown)


includes/class-wc-google-gtag-js.php
```diff
@@ -86,27 +86,86 @@ public function variant_data( $data, $product, $variation ) {
...
+			'view_item_list',
+			array(
+				'items' => array(
+					array(
+						'id'            => $product->get_id(),
+						'name'          => $product->get_title(),
+						'category'      => self::product_get_category_line( $product ),
+						// @codingStandardsIgnoreStart
+						'list'          => isset( $_GET['s'] ) ? __( 'Search Results', 'woocommerce-google-analytics-integration' ) : __( 'Product List', 'woocommerce-google-analytics-integration' ),
+						// @codingStandardsIgnoreEnd
```
mikkamp: I wouldn't recommend to turn off all phpcs rules here. We generally also use the format like this:
`// phpcs:ignore Squiz.Arrays.ArrayDeclaration.SingleLineNotAllowed`

In my VSCode I have the option "Show sniff source codes in diagnostic messages" enabled which means I can just copy and paste the specific sniff code to ignore.

```suggestion
						// phpcs:ignore WordPress.Security.NonceVerification.Recommended
						'list'          => isset( $_GET['s'] ) ? __( 'Search Results', 'woocommerce-google-analytics-integration' ) : __( 'Product List', 'woocommerce-google-analytics-integration' ),
```

In this case it's important to ignore a specific rule, because we still want it to check that the right i18n domain name is being used. Or any other rules for that line.

includes/class-wc-google-gtag-js.php
```diff
@@ -201,26 +260,26 @@ function gtag(){dataLayer.push(arguments);}
...
-		// Order items
-		$items = "[";
-		if ( $order->get_items() ) {
-			foreach ( $order->get_items() as $item ) {
-				$items .= self::add_item( $order, $item );
+		$event_items = array();
+		$order_items = $order->get_items();
+		if ( ! empty( $order_items ) ) {
+			foreach ( $order_items as $item ) {
+				$event_items += self::add_item( $order, $item );
```
mikkamp: I don't think this is intended here. The function `add_item` returns an array (would be nice to mention that in the docblock). So when we combine two arrays with the `+` operator we essentially do a union of the two arrays. You compensate for that later by wrapping `event_items` in another array. Which will match up as long as you are tracking only one order item, but as soon as you add multiple order items that's going to fail.
```suggestion
				$event_items[] = self::add_item( $order, $item );
```

I'd suggest to try running this code to see how that differs:

```php
$one = [ 'id' => 1, 'label' => 'One' ];
$two = [ 'id' => 2, 'label' => 'Two' ];

$numbers =  [];
$numbers += $one;
$numbers += $two;

var_dump( array( $numbers ) );

$numbers   = [];
$numbers[] = $one;
$numbers[] = $two;

var_dump( $numbers );
```

includes/class-wc-google-gtag-js.php
```diff
@@ -201,26 +260,26 @@ function gtag(){dataLayer.push(arguments);}
...
+		return self::get_event_code(
+			'purchase',
+			array(
+				'transaction_id' => $order->get_order_number(),
+				'affiliation'    => get_bloginfo( 'name' ),
+				'value'          => $order->get_total(),
+				'tax'            => $order->get_total_tax(),
+				'shipping'       => $order->get_total_shipping(),
+				'currency'       => $order->get_currency(),
+				'items'          => array( $event_items ),
```
mikkamp: As mentioned above this should not be wrapped in an array.

```suggestion
				'items'          => $event_items,
```

includes/class-wc-google-gtag-js.php
```diff
@@ -86,27 +86,86 @@ public function variant_data( $data, $product, $variation ) {
...
+	 * @return array
+	 */
+	public static function product_get_category_line( $_product ) {
+		$out            = [];
+		$variation_data = $_product->is_type( 'variation' ) ? wc_get_product_variation_attributes( $_product->get_id() ) : false;
+		$categories     = get_the_terms( $_product->get_id(), 'product_cat' );
+
+		if ( is_array( $variation_data ) && ! empty( $variation_data ) ) {
+			$parent_product = wc_get_product( $_product->get_parent_id() );
+			$categories     = get_the_terms( $parent_product->get_id(), 'product_cat' );
```
mikkamp: Loading the parent product can be useful to ensure it's actually a valid product, but in this case we only use it to get the ID, which we already have. For example if `wc_get_product` would return false you'd get a fatal error here. Maybe the following would provide better fallbacks?

```suggestion
			$categories = get_the_terms( $_product->get_parent_id(), 'product_cat' );
```

includes/class-wc-google-gtag-js.php
```diff
@@ -86,27 +86,86 @@ public function variant_data( $data, $product, $variation ) {
...
+		}
+
+		if ( $categories ) {
```
mikkamp: Might want to double check that it's an array or use `! is_wp_error`, because `get_the_terms` from the main product or the parent product could return a WP_Error which would cause the `foreach` to fail.

includes/class-wc-google-gtag-js.php
```diff
@@ -86,27 +86,86 @@ public function variant_data( $data, $product, $variation ) {
...
+	}
+
+	/**
+	 * Returns an array of category names the product is atttributed to
+	 *
+	 * @param  WC_Product $_product  Product to pull info for
+	 * @return array
+	 */
+	public static function product_get_category_line( $_product ) {
+		$out            = [];
```
mikkamp: Happy to be using short array syntax, but I do think it's nicer to be consistent. Everywhere else in this PR you use the long syntax.

includes/class-wc-google-gtag-js.php
```diff
@@ -296,37 +367,34 @@ public function product_detail( $product ) {
...
-				$items .= "
-					'variant': " . $variant;
+				$item_data['variant'] = $variant;
 			}
 
-			$items .= "
-					'price': '" . esc_js( $product->get_price() ) . "',
-					'quantity': '" . esc_js( $cart_item['quantity'] ) . "'
-				},";
+			$items[] += $item_data;
```
mikkamp: If we are just appending to the array we don't need the `+` operator.

```suggestion
			$items[] = $item_data;
```

mikkamp (commented):
Thanks for the changes, the snippets are pretty much showing on the page as they were previously. However I was getting an error on the single product page.

The add to cart category is missing single quotes and a comma. I had to change `class-wc-google-analytics.php:496` to the following code:

```php
$item .= "'category': '" . $this->get_tracking_instance()->product_get_category_line( $product ) . "', ";
```

After that the tracking snippets started working. I saw some unit tests you did in another branch, and was hoping those were included here so I wouldn't have to check each event manually. I assume those are still coming in a follow up PR?

I also left a few other comments as I went through the code.

includes/class-wc-google-gtag-js.php
martynmjones: Thanks for the suggestion Mik - I've updated the [rule being ignored here](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2b656ac996cdf886583977a39c77a31d7487db39/includes/class-wc-google-gtag-js.php#L158).

includes/class-wc-google-gtag-js.php
martynmjones: I've reworked the `product_get_category_line` [function](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2b656ac996cdf886583977a39c77a31d7487db39/includes/class-wc-google-gtag-js.php#L125) instead of having a straight copy from the abstract analytics class to address the points raised.

martynmjones:
Thanks for the review here @mikkamp 👍 I've pushed changes to address the points you raised with the biggest changes being that the gtag class has it's own [add_to_cart](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2b656ac996cdf886583977a39c77a31d7487db39/includes/class-wc-google-gtag-js.php#L214) function and [loop_add_to_cart](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2b656ac996cdf886583977a39c77a31d7487db39/includes/class-wc-google-analytics.php#L555) is no longer being used with gtag as the product archive add to cart event is now being handled by the [listing_click](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2b656ac996cdf886583977a39c77a31d7487db39/includes/class-wc-google-gtag-js.php#L175) function.

includes/class-wc-google-analytics.php
```diff
@@ -475,35 +475,39 @@ public function add_to_cart() {
...
 		global $product;
 
-		// Add single quotes to allow jQuery to be substituted into _trackEvent parameters
-		$parameters = array();
-		$parameters['category'] = "'" . __( 'Products', 'woocommerce-google-analytics-integration' ) . "'";
-		$parameters['action']   = "'" . __( 'Add to Cart', 'woocommerce-google-analytics-integration' ) . "'";
-		$parameters['label']    = "'" . esc_js( $product->get_sku() ? __( 'ID:', 'woocommerce-google-analytics-integration' ) . ' ' . $product->get_sku() : "#" . $product->get_id() ) . "'";
+		if ( 'yes' === $this->ga_gtag_enabled ) {
+			$this->get_tracking_instance()->add_to_cart( $product );
+		} else {
```
mikkamp: I personally prefer to reduce indented code blocks whenever possible. For example in this case the sole action is to run the class function instead and then exit the function, so if we add a return then we no longer need to move everything else in an else block.

```suggestion
		if ( 'yes' === $this->ga_gtag_enabled ) {
			$this->get_tracking_instance()->add_to_cart( $product );
			return;
		}
```

includes/class-wc-google-analytics.php
```diff
@@ -475,35 +475,39 @@ public function add_to_cart() {
...
+				$item .= "'quantity': $( 'input.qty' ).val() ? $( 'input.qty' ).val() : '1'";
+				$item .= '}';
 
-			$parameters['item'] = $item;
+				$parameters['item'] = $item;
 
-			$code = "" . $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . " );";
-			$parameters['enhanced'] = $code;
-		}
+				$code                   = $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', " . $item . ' );';
```
mikkamp: Since we are already using double quotes and a single variable we can just interpolate instead of concat. In my opinion slightly easier to read.

 ```suggestion
				$code                   = $this->get_tracking_instance()->tracker_var() . "( 'ec:addProduct', {$item} );";
```

File: includes/class-wc-google-gtag-js.php [hunk 10]
```diff
@@ -86,58 +86,166 @@ public function variant_data( $data, $product, $variation ) {
...
+			item_data.quantity = $("input.qty").val() ? $("input.qty").val() : 1;';
+
+		if ( $product->is_type( 'variable' ) ) {
+			// Check the global google_analytics_integration_product_data Javascript variable contains data
+			// for the current variation selection and if it does update the item_data to be sent for this event
+			$event_code .= "
+			if ( google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ] !== undefined ) {
+				item_data.id       = google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].id;
+				item_data.variant  = google_analytics_integration_product_data[ $('input[name=\"variation_id\"]').val() ].variant;
+			}
```
mikkamp: I'm not sure if jQuery applies any caching but from this code it looks like we are asking it to do the following 3 times in a row:
- Search for an input on the page
- Check if it matches the specific variation
- Retrieve the value of the input

Seems like we could just do that once and store it in a `const`. It would also reduce the need to escape double quotes which makes the code block harder to read.

includes/class-wc-google-gtag-js.php
```diff
@@ -233,36 +341,42 @@ protected function add_item( $order, $item ) {
 		$_product = $item->get_product();
 		$variant  = self::product_get_variant_line( $_product );
 
-		$code  = '{';
-		$code .= "'id': '" . self::get_product_identifier( $_product ) . "',";
-		$code .= "'name': '" . esc_js( $item['name'] ) . "',";
-		$code .= "'category': " . self::product_get_category_line( $_product );
+		$event_item = array(
+			'id'       => self::get_product_identifier( $_product ),
```
mikkamp: I know it wasn't introduced in this PR, but we don't really need to use an underscore prefix especially since it will remain within the scope of this function. So I'd stick with just `$product`.

mikkamp (commented):
Thanks for the changes and the refactoring. It seems to be working as intended now, I'm seeing the various events getting tracked.

I just left a few more minor comments as I looked through the code.

includes/class-wc-google-analytics.php
martynmjones: Thanks @mikkamp - I've [adjusted the indentation](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/240/commits/e44bb50b01307a9249a5e914405fd11d068b8c2a) as you suggested.

includes/class-wc-google-gtag-js.php
martynmjones: You're absolutely right that it's not particularly reader-friendly at the moment! I've [adjusted the code](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/240/commits/843e13ddca1651689048ac228351ea830572f0ef) and it looks a lot better now.

includes/class-wc-google-analytics.php
martynmjones: I've [updated](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/240/commits/859677a7cc31b488bdb249c041a9cfe03c2a969f) that line of code with your suggestion.

includes/class-wc-google-gtag-js.php
martynmjones: Very true! I've [updated the variable name](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/5675be7a712ee0912e5138476c87e8bcea86b165/includes/class-wc-google-gtag-js.php#L342) to be more in keeping with how things are done elsewhere.

mikkamp:
Thanks for the additional changes, that looks like some cleaner code.

I was testing the changes in detail (previously I mainly looked at the JS code snippets and how they were added to the page), in particular I'm looking at the data which gets sent to Google Analytics. I'm getting a bit of mixed results though, and I feel there are some parts missing.

## Variable product `add_to_cart`
For example when I click on a variable product and add it to the cart I'm seeing the `add_to_cart` event getting sent. But it seems to be missing the product details:
image

On the same page if I click on one of the single related products I do so the product details included in the `add_to_cart` event:
image

I'm looking at the code snippet and I assume that items should be an array containing `data_item`:
```js
gtag('event', 'add_to_cart', {"items": item_data})
```

Changing it to an array seems to work:
image

## Cart page `remove_from_cart`
I tested this event and it seems like it's sending jQuery code instead of the values:
image

Looking at the matching JS snippet it seems it's adding double quotes around sections it shouldn't:
```
<script>
function($) {
    $( '.remove' ).off('click', '.remove').on( 'click', function() {
        gtag('event', 'remove_from_cart', {"items":[{"id":"($(this).data(\\'product_sku\\')) ? ($(this).data(\\'product_sku\\')) : (\\'#\\' + $(this).data(\\'product_id\\'))","quantity":"$(this).parent().parent().find( \\'.qty\\' ).val() ? $(this).parent().parent().find( \\'.qty\\' ).val() : \\'1\\',"}]})
    });
})(jQuery);
</script>
```

## Archive page `select_content`
This is triggered from any product that doesn't have a direct add to cart button (tested with a grouped product). I'm seeing very little information get included:
image

It seems it's sending a list of items (same as add_to_cart buttons) but that doesn't go through since according to the docs it does not accept that: https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#select_content
Maybe we can log an issue, since it seems to already have been done that way before this PR.

## Events `view_item_list` and `view_item`
Contain the information I expect:
```
en=view_item_list&_ee=1&pr1=id43~nmGrouped%20Product~caUncategorized~lnProduct%20List~lp3&ep.link_attribution=true&ep.anonymize_ip=true&ep.logged_in=yes&_et=1
en=view_item&_ee=1&pr1=id%2385~nmVariable~caUncategorized~pr2&ep.link_attribution=true&ep.anonymize_ip=true&ep.logged_in=yes
```

## Event `begin_checkout` is working as expected
```
en=begin_checkout&_ee=1&pr1=id%2331~nmHamburg%20K%C3%BChlschrankmagnet%20Landungsbr%C3%BCcken~caUncategorized~pr10~qt4&pr2=id%2385~nmVariable~caUncategorized~pr2~qt2~va%5C'Color%3A%20Blue%5C'%2C&pr3=id%2385~nmVariable~caUncategorized~pr4~qt1~va%5C'Color%3A%20Green%5C'%2C&ep.link_attribution=true&ep.anonymize_ip=true&ep.logged_in=yes&_et=2
```

## Event `purchase` is working as expected
```
en=purchase&_ee=1&pr1=id%2319~nmProduct%20Number%20Eight~caMain~pr80~qt2&ep.link_attribution=true&ep.anonymize_ip=true&ep.logged_in=yes&ep.transaction_id=98&ep.affiliation=woo-analytics.test&ep.value=170.00&ep.tax=0&ep.shipping=10&_et=1&cu=EUR
```

martynmjones:
Thanks for the detailed testing @mikkamp!

The `remove_from_cart` event has been [updated so that the script executes](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/240/commits/eebfd3996bd5b6ef396057928627ae28ff00d874) and the `add_to_cart` event item data is [now an array](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/240/commits/054f700ebb73331027883262812a027465b7293f) so both of these should now be working as expected.

As you mentioned the issue with the `select_content` event has been in place for a while so I've opened a new issue for this #243 to address it separately.

mikkamp (approved):
Thanks for those changes the `add_to_cart` is working now. The `remove_from_cart` has improved although once again `items` should be an array of objects, not a single object.

Besides that it's all working fine, so I'll go ahead and approve the PR already.

PR #245, opened 2023-01-12
Add PHP unit tests

martynmjones:

This PR adds basic unit testing for any events making use of the `woocommerce_gtag_event_data` filter. That doesn't include events that rely on Javascript like `add_to_cart`, however, it gives us a starting point for tests.

Testing can then be revisited in the future when the core WooCommerce E2E testing implementation has been updated so that coverage can be expanded.

### Checks:

### Screenshots:

Screenshot 2023-01-12 at 17 04 45

_The PHPUnit implementation is based on how AutomateWoo tests work._

From the plugin directory, checkout `add/php-unit-tests` and then run:

1. `composer install`
2. `bin/install-unit-tests.sh <db-name> <db-user> <db-pass> [db-host] latest latest`
3. `vendor/bin/phpunit`

### DIFF
--- a/bin/unit-tests-functions.sh
+++ b/bin/unit-tests-functions.sh
@@ -119,7 +119,7 @@ install_test_suite() {
     local WP_TESTS_TAG="tags/$LATEST_VERSION"
   fi
 
-  echo "Install test suit ${WP_TESTS_TAG}"
+  echo "Install test suite ${WP_TESTS_TAG}"
 
   # portable in-place argument for both GNU sed and Mac OSX sed
   if [[ $(uname -s) == 'Darwin' ]]; then
--- a/bin/unit-tests-functions.sh
+++ b/bin/unit-tests-functions.sh
@@ -1,0 +1,256 @@
+#!/usr/bin/env bash
+
+TMPDIR=${TMPDIR-/tmp}
+TMPDIR=$(echo $TMPDIR | sed -e "s/\/$//")
+WP_TESTS_DIR=${WP_TESTS_DIR-$TMPDIR/wordpress-tests-lib}
+WP_CORE_DIR=${WP_CORE_DIR-$TMPDIR/wordpress}
+
+# Plugin variables.
+PLUGINS_DIR="${WP_CORE_DIR}/wp-content/plugins"
+
+download() {
+  if [ $(which curl) ]; then
+    curl -s "$1" >"$2"
+  elif [ $(which wget) ]; then
+    wget -nv -O "$2" "$1"
+  fi
+}
+
+get_latest_wp_version() {
+  # http serves a single offer, whereas https serves multiple. we only want one
+  download http://api.wordpress.org/core/version-check/1.7/ "${TMPDIR}/wp-latest.json"
+  local LATEST_VERSION=$(grep -o '"version":"[^"]*' "${TMPDIR}/wp-latest.json" | sed 's/"version":"//')
+  if [[ -z "$LATEST_VERSION" ]]; then
+    echo "Latest WordPress version could not be found"
+    exit 1
+  fi
+  echo $LATEST_VERSION
+}
+
+cleanup() {
+  if [ -d "$WP_CORE_DIR" ]; then
+    rm -rf "$WP_CORE_DIR"
+  fi
+
+  if [ -d "$WP_TESTS_DIR" ]; then
+    rm -rf "$WP_TESTS_DIR"
+  fi
+}
+
+install_wp() {
+  local WP_VERSION=$1
+
+  if [[ $WP_VERSION == 'nightly' || $WP_VERSION == 'trunk' ]]; then
+    echo "Installing WordPress ($WP_VERSION)."
+    cleanup
+    mkdir -p "$WP_CORE_DIR"
+    mkdir -p "$TMPDIR"/wordpress-trunk
+    rm -rf "$TMPDIR"/wordpress-trunk/*
+    svn export --quiet https://core.svn.wordpress.org/trunk "$TMPDIR"/wordpress-trunk/wordpress
+    mv "$TMPDIR"/wordpress-trunk/wordpress/* "$WP_CORE_DIR"
+  else
+    if [ $WP_VERSION == 'latest' ]; then
+      local ARCHIVE_NAME='latest'
+      local LATEST_VERSION=$(get_latest_wp_version)
+    elif [[ $WP_VERSION =~ [0-9]+\.[0-9]+ ]]; then
+      # https serves multiple offers, whereas http serves single.
+      download https://api.wordpress.org/core/version-check/1.7/ "$TMPDIR"/wp-latest.json
+      if [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0] ]]; then
+        # version x.x.0 means the first release of the major version, so strip off the .0 and download version x.x
+        local LATEST_VERSION=${WP_VERSION%??}
+      else
+        # otherwise, scan the releases and get the most up to date minor version of the major release
+        local VERSION_ESCAPED=$(echo $WP_VERSION | sed 's/\./\\./g')
+        local LATEST_VERSION=$(grep -o '"version":"'$VERSION_ESCAPED'[^"]*' "$TMPDIR"/wp-latest.json | sed 's/"version":"//' | head -1)
+      fi
+      if [[ -z "$LATEST_VERSION" ]]; then
+        local ARCHIVE_NAME="wordpress-$WP_VERSION"
+      else
+        local ARCHIVE_NAME="wordpress-$LATEST_VERSION"
+      fi
+    else
+      local ARCHIVE_NAME="wordpress-$WP_VERSION"
+    fi
+
+    local TARGET_VERSION=${LATEST_VERSION:-$WP_VERSION}
+    local WP_VERSION_FILE="${WP_CORE_DIR}/version-"$(echo $TARGET_VERSION | sed -e "s/\//-/")
+
+    if [ -f "$WP_VERSION_FILE" ]; then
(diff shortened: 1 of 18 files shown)


bin/unit-tests-functions.sh
```diff
@@ -0,0 +1,256 @@
...
+      local WP_TESTS_TAG="tags/$WP_VERSION"
+    fi
+  elif [[ $WP_VERSION == 'nightly' || $WP_VERSION == 'trunk' ]]; then
+    local WP_TESTS_TAG="trunk"
+  else
+    local LATEST_VERSION=$(get_latest_wp_version)
+    local WP_TESTS_TAG="tags/$LATEST_VERSION"
+  fi
+
+  echo "Install test suit ${WP_TESTS_TAG}"
```
mikkamp: I know this is just copied, but this should be `suite`

includes/class-wc-google-gtag-js.php
```diff
@@ -140,6 +140,15 @@ public static function product_get_category_line( $product ) {
 		return join( '/', $category_names );
 	}
 
+	/**
+	 * Return list name for event
+	 *
+	 * @return string
+	 */
+	public static function get_list_name(): string {
+		return isset( $_GET['s'] ) ? __( 'Search Results', 'woocommerce-google-analytics-integration' ) : __( 'Product List', 'woocommerce-google-analytics-integration' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
mikkamp: This is already a long line. We can put the `phpcs:ignore` on the line above (which means it will ignore the following line. Makes it just slightly easier to read.

phpunit.xml.dist
```diff
@@ -0,0 +1,21 @@
...
+			<directory suffix=".php">./tests/unit-tests</directory>
+		</testsuite>
+	</testsuites>
+	<coverage>
+		<include>
+			<directory suffix=".php">./includes</directory>
+			<directory suffix=".php">./admin</directory>
+			<file>woocommerce-google-analytics-integration.php</file>
+		</include>
+	</coverage>
```
mikkamp, line 20: Nice to see this section included, now with phpunit 9.5 (and xdebug configured for PHP) it becomes much easier to get a code coverage report.

Works for me by running:
```
XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-text
```
image


Can we get those instructions added to the README.md file? It would be nice to add the whole section about running the unit tests: https://github.com/woocommerce/automatewoo/blob/develop/README.md#unit-tests
Might as well copy the section for PHPCS while we are at it.

tests/unit-tests/AddTransactionEnhanced.php
```diff
@@ -0,0 +1,58 @@
+<?php
+// phpcs:ignoreFile
```
mikkamp: I don't think we should promote adding this ignore file in each test. If we really don't want these files to be checked then we can exclude the whole tests folder in phpcs.xml.dist. Although I much prefer the approach we took here: https://github.com/woocommerce/google-listings-and-ads/pull/1624

That enforces good styling in tests but doesn't get pedantic about needing doc blocks for each test function. Your tests already conform to good styling so it's just the rules for the docblocks that will allow it to fully pass.

tests/unit-tests/ListingImpression.php
```diff
@@ -0,0 +1,47 @@
...
+		$position = 1;
+
+		// Mock woocommerce_gtag_event_data filter to ensure it is called and the correct data is processed
+		$filter = new MockAction();
+		add_filter( 'woocommerce_gtag_event_data', array( &$filter, 'filter' ) );
+
+		( new WC_Google_Gtag_JS() )->listing_impression( $product, $position );
+
+		// Confirm woocommerce_gtag_event_data is called by listing_impression()
+		$this->assertEquals( 1, $filter->get_call_count(), 'woocommerce_gtag_event_data filter was not called for view_item_list (listing_impression()) event' );
```
mikkamp: There is quite a lot of common code to test an event, such as setting up a filter, confirm the filter was called, create an instance of WC_Google_Gtag_JS, check that the expected data matches the filter arguments.

I'd see two ways to simplify event testing:

1. Create a common event testing class (similar to how GLA creates [these TestCase classes](https://github.com/woocommerce/google-listings-and-ads/tree/develop/tests/Framework)):
```php
class GtagEventTestCase extends WP_UnitTestCase {
```

Which can contain some common methods for setting up the filter / checking it, or alternatively add it to the `setUp` function.

2. Test all the events in one class
Have a setUp function which runs some common code for each test.

tests/class-unittestsbootstrap.php
```diff
@@ -0,0 +1,153 @@
...
+		echo 'WooCommerce Finished Installing...' . PHP_EOL;
+
+	}
+
+	/**
+	 * Load test cases and factories
+	 *
+	 * @since 2.0
+	 */
+	public function includes() {
```
mikkamp, line 136: One thing I dislike about AW is that it doesn't autoload it's own Test Cases / helpers and they have to be added to this include function manually. I think it's ok to load the WC ones manually but the composer autoloader can be used for any framework files we add in tests (like is done in GLA).

> **Note**
You don't have any framework files yet, but one of the comments I made suggests to add some common code there, so this only applicable if you decide to go with implementing that.

tests/unit-tests/CheckoutProcess.php
```diff
@@ -0,0 +1,75 @@
...
+		( new WC_Google_Gtag_JS() )->checkout_process( $cart->get_cart() );
+
+		// Confirm woocommerce_gtag_event_data is called by checkout_process()
+		$this->assertEquals( 1, $filter->get_call_count(), 'woocommerce_gtag_event_data filter was not called for begin_checkout (checkout_process()) event' );
+
+		// The expected data structure for this event
+		$expected_data = array(
+			'items' => array(
+				array(
+					'id'       => WC_Google_Gtag_JS::get_product_identifier( $product ),
```
mikkamp: I have mixed feelings about calling these helper function to generate the test data. As it is now if there is a bug in `get_product_identifier` where it returns the ID as `#*123` then the test will still pass because they both are calling the buggy function.

But then again the purpose of this test is to confirm the array structure is correct, we'd need separate unit tests which would confirm that the `get_product_identifier` function returns the right data. Since those tests are out of scope for this PR I guess we could leave it as is. Since it's more about failure to fully cover functions in tests.

mikkamp (commented):
Thanks for creating this PR, when I run it the tests pass without issues. I'm able to run it both on PHP 7.4 and 8.0. 

I just left some comments as I went through the code changes.

Also are we getting a followup PR to automatically test in GitHub Actions?

phpunit.xml.dist
martynmjones: Thanks for the suggestion @mikkamp - I've updated the README.md file to include [testing instructions along with code coverage](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/add/php-unit-tests/README.md).

tests/unit-tests/AddTransactionEnhanced.php
martynmjones: I've removed the ignore rule from all the tests and added required comments to pass standards.

tests/unit-tests/ListingImpression.php
martynmjones: Thanks again for the recommendations here - I went with option 1 and have added a [common class](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/add/php-unit-tests/tests/UnitTest.php) that removes some of the repetition from the individual unit tests.

tests/class-unittestsbootstrap.php
martynmjones: The framework file is now being autoloaded using Composer to avoid manually including it - https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/245/commits/1a6017d19bda17270bd816d3a6a14fefcfd7613a

tests/unit-tests/CheckoutProcess.php
martynmjones: I agree we need to expand test coverage. For now my focus, like you said, was purely on testing the data structure for events. I've opened issue #248 so that we can look to expand coverage down the road.

tests/unit-tests/AddTransactionEnhanced.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+class AddTransactionEnhanced extends UnitTest {
+
+	/**
+	 * Run unit test against the `purchase` event
+	 *
+	 * @return void
+	 */
+	public function test_purchase_event() {
+		$order = parent::get_order();
```
mikkamp: I personally think it would be a bit more common to call `$this->get_order()` since we are extending the class. Although I myself would have approached it slightly different.

Right now it will create a product, order and customer, since it checks null it looks like it will keep the same product for the whole class, regardless if it changes or not. Commonly we'll have one test with a product with specific data and then a next one with a slightly different product. The `UnitTest` class model wouldn't be conducive for those kinds of tests.
Not necessarily saying that this should be changed, maybe it would just be clearer to give it a bit more specific name. `UnitTest` gives the implication that we should extend it for every unit test, but it's built specifically for aiding in testing an event.

mikkamp (approved):
Thanks for the changes, the tests are passing and I think it looks a bit less redundant now. I'll go ahead and approve this PR.

I just left some comments about the UnitTest class, might be better to clarify the reusability for future tests we add.

tests/unit-tests/AddTransactionEnhanced.php
martynmjones: Thanks for the suggestion. I've [renamed](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/245/commits/d0416f8e709b803482e83e6cec0a9b4472048a9c) the class so it's a bit clearer.

PR #246, opened 2023-01-13
Fix usage of tracker_var() in load_analytics()

### LINKED ISSUE #185: Inconsistent usage of woocommerce_ga_tracker_variable

martynmjones:

In some places, the code in `WC_Google_Gtag_JS::load_analytics()` was directly printing `gtag()` when setting up the tracking instance. This PR adjusts the code so that `tracker_var()` is used in its place to ensure the filter `woocommerce_gtag_tracker_variable` is applied consistently.

Closes #185.

### Checks:

### Screenshots:

Screenshot 2023-01-13 at 06 35 17

Screenshot 2023-01-13 at 06 38 50

1. Checkout `fix/185-usage-of-tracker_var-in-setup` on a test site that has the extension active and set up
2. Use the `woocommerce_gtag_tracker_variable` filter to alter the default tracker var
_Example:_
```
add_filter(
	'woocommerce_gtag_tracker_variable',
	function( $tracker_var ) {
		return 'filteredGtag';
	}
);
```
3. View source of the frontend of your site and search for the inline script output by [load_analytics()](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/86193d93b4973c03269b7ae603e55264dca8887c/includes/class-wc-google-gtag-js.php#L258)
4. Confirm that all instances of `gtag()` have been replaced with the filtered tracker var

### DIFF
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -273,11 +273,11 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 		$gtag_cross_domains = ! empty( self::get( 'ga_linker_cross_domains' ) ) ? array_map( 'esc_js', explode( ',', self::get( 'ga_linker_cross_domains' ) ) ) : array();
 
 		$gtag_snippet = '<script async src="https://www.googletagmanager.com/gtag/js?id=' . esc_js( $gtag_id ) . '"></script>';
-		$gtag_snippet .= "
+		$gtag_snippet .= '
 		<script>
 		window.dataLayer = window.dataLayer || [];
-		function " . self::tracker_var() . "(){dataLayer.push(arguments);}
-		" . self::tracker_var() . "('js', new Date());
+		function ' . self::tracker_var() . '(){dataLayer.push(arguments);}
+		' . self::tracker_var() . "('js', new Date());
 		$gtag_developer_id
 
 		" . self::tracker_var() . "('config', '" . esc_js( $gtag_id ) . "', {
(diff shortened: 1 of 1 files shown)


includes/class-wc-google-gtag-js.php
```diff
@@ -476,3 +476,10 @@ public function event_tracking_code( $parameters, $selector ) {
 	}
 
 }
+
+add_filter(
+	'woocommerce_gtag_tracker_variable',
+	function( $tracker_var ) {
+		return 'filteredGtag';
+	}
+);
```
mikkamp: I assume this was just for testing and not intended to end up in the final PR.

mikkamp (approved):
Thanks, it looks good and is working as expected. I can't find any other instances where the helper function isn't used.

Since the usage is 4 times within the function I would have contemplated using a variable and then use it in the string as `{$tracker_var}`, that would prevent having to switch back and forth between single and double quotes.

I'll go ahead and approve it but it just needs the test filter snippet removed.

includes/class-wc-google-gtag-js.php
martynmjones: 🤦 Thanks for picking up on that @mikkamp that was absolutely a mistake. I've removed the stray filter 👍

PR #260, opened 2023-03-15
Refactor wp_head scripts and use wp_register scripts instead.

### LINKED ISSUE #186: Analytics Scripts should use the WordPress Scripts API

### LINKED ISSUE #196: Remove text/javascript from all scripts
Hello.
Can you remove the text/javascript declaration from class-wc-abstract-google-analytics-js.php from line 51 or enqueue it with core’s inline script actions? According to HTML5 and html validator
“The type attribute is unnecessary for JavaScript resources”.

You enqueue the above script by echoing it, so the below function cannot affect it
```
   'after_setup_theme',
    function() {
        add_theme_support( 'html5', [ 'script', 'style' ] );
    }
```
Same applies to class-wc-google-analytics-js.php at 104, 175 and class-wc-google-analytics.php 414

A workaround could be something like the below:

```
$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
?>
<script<?php echo $type_attr; ?>>
// ....
</script>
```

### LINKED ISSUE #212: When used with WC google-analytics-integration, the Google Analytics script is loaded twice
Hello, 
i have https://github.com/woocommerce/woocommerce-google-analytics-integration
and now with your plugin, i have two requests to google analytics :
Screenshot from 2021-12-03 15-56-08

it seems that 
```
if ( ! wp_script_is( 'google-tag-manager', 'registered' ) ) {
```
is a great check but it is missing in woocommerce-google-analytics-integration , which comes after on my screenshot.
could you talk to each other and make both script from the same corporation behave together ?

thank you !

puntope:
Closes #212 
Closes #186 
Closes #196 

This PR implements wp_register_scripts, and wc_enqueue_js avoiding conflicts with other plugins loading google tag manager and google analytics.

### Checks:

1. Successfully run [all the smoke tests in the wiki](https://github.com/woocommerce/woocommerce-google-analytics-integration/wiki/General-Testing)
2. Check the code to see if makes sense
3. Be sure there are no duplicate google tag manager anymore. 

### Additional details:
⚠️  We removed [includes/class-wc-google-analytics-info-banner.php](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/files/5349e019787ce63177ea0f8125681a03a2f98637#diff-da1beddc018c0eb5a8fd40b0614f0c3c8831640ac40455ff2211f57ef2489bfe) since it's not being used anymore. 

### Changelog entry

> Fix - Avoid duplication of Google Tag Manager script
> Dev - Load scripts via `wp_register_scripts` and `wp_eneuque_js`
> Dev - Refactor files allowing PHPCS to pass


includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -1,4 +1,8 @@
 <?php
+
+// @phpcs:disable Squiz.Classes.ClassFileName.NoMatch
+// @phpcs:disable Squiz.Classes.ValidClassName.NotCamelCaps
```
mikkamp: We explicitly enable all `Squiz.Classes` rules here (with two exceptions): https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/1.6.2/phpcs.xml.dist#L44-L47

Which seems to be copied from GLA as we don't do that in every extension. So either I think we should remove the enable or add the exclusion to the config file, and not do it per file.

includes/class-wc-google-analytics-info-banner.php
```diff
@@ -1,114 +0,0 @@
-<?php
```
mikkamp, line 1: Curious why this file is getting removed. I can see in trunk that it's no longer referenced, but I don't see a specific reasoning why it was removed.

Looks like it was removed in this PR #235. Would be nice to mention it in the PR description since it's not related to the changes.

includes/class-wc-google-analytics-js.php
```diff
@@ -1,4 +1,8 @@
 <?php
+
+// @phpcs:disable Squiz.Classes.ClassFileName.NoMatch
+// @phpcs:disable Squiz.Classes.ValidClassName.NotCamelCaps
+
```
mikkamp: Again same as above, preferable to configure this for all files at once.

mikkamp (commented):
Got started on the review but ran out of time. Just wanted to submit comments so they don't get lost, I'll continue the review tomorrow.

includes/class-wc-google-analytics-js.php
```diff
@@ -172,19 +180,21 @@ public static function classic_analytics_footer() {
...
+		wc_enqueue_js( $code );
 	}
 
 	/**
 	 * Enqueues JavaScript to send the pageview last thing (needed for things like addImpression)
 	 */
 	public static function universal_analytics_footer() {
 		if ( apply_filters( 'wc_google_analytics_send_pageview', true ) ) {
-			wc_enqueue_js( "" . self::tracker_var() . "( 'send', 'pageview' ); " );
+			wc_enqueue_js( '' . self::tracker_var() . "( 'send', 'pageview' ); " );
```
mikkamp: Prefixing with an empty string is not necessary, it's an old trick to convert the result of `self::tracker_var()` to a string. Much better to typecast to string in the `tracker_var` function. Repeated several times in this file.

includes/class-wc-google-analytics-js.php
```diff
@@ -491,41 +500,45 @@ public function checkout_process( $cart ) {
...
 		if ( 'yes' === self::get( 'ga_use_universal_analytics' ) ) {
 			if ( 'yes' === self::get( 'ga_enhanced_ecommerce_tracking_enabled' ) ) {
-				wc_enqueue_js( "
+				wc_enqueue_js(
+					"
 					$( '" . $selector . "' ).on( 'click', function() {
-						" . $parameters['enhanced'] . "
-						" . self::tracker_var() . "( 'ec:setAction', 'add' );
+						" . $parameters['enhanced'] . '
+						' . self::tracker_var() . "( 'ec:setAction', 'add' );
```
mikkamp, line 517: Since we are using static we can't interpolate using `{$this->tracker_var()}` but if it's used many times in a function we can make it a bit cleaner by doing something like:

```php
$tracker_var = self::tracker_var();

$code = "
    {$tracker_var}( 'ec:setAction', 'add' );
    {$tracker_var}( 'send', 'event', 'UX', 'click', 'add to cart' );
";
```

Makes the code snippets easier to read so we don't have to go and open/close quotes all the time. Ideally we wouldn't have to be dealing with all these static functions, and the tracker var could have been initialised once as a class member, so we don't have to call apply_filters over and over again, but that's probably a bigger refactor left for another day.

includes/class-wc-google-analytics.php
```diff
@@ -388,66 +396,56 @@ public function tracking_code_display() {
...
 
 		if ( is_woocommerce() || is_cart() || ( is_checkout() && ! $display_ecommerce_tracking ) ) {
 			$display_ecommerce_tracking = true;
-			echo $this->get_standard_tracking_code();
+			$this->get_standard_tracking_code();
 		}
 
 		if ( ! $display_ecommerce_tracking && 'yes' === $this->ga_standard_tracking_enabled ) {
-			echo $this->get_standard_tracking_code();
+			$this->get_standard_tracking_code();
```
mikkamp: We are no longer echoing all these functions, because they are no longer returning the tracking code but enqueing instead. Can we rename these functions to clarify the change in function. Like:
`enqueue_standard_tracking_code`
`enqueue_ecommerce_tracking_code`

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -57,45 +61,44 @@ public static function header() {
 			function gaOptout() {
 				document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
 				window[disableStr] = true;
-			}
-		</script>";
+			}";
+
+		wc_enqueue_js( $code );
```
mikkamp: This code used to be included in the header right before the general analytics code, but it's now moved to the footer, where it's probably too late to work as an opt out script.

Is it better to use a combination of `wp_register_script` and `wp_add_inline_script` here and give it a new handle like `google-analytics-opt-out`? When we call `wp_register_script` for `google_tag_manager` we can then add `google-analytics-opt-out` as a dependency so it ensures it's output on the page first.

```suggestion
		wp_register_script( 'google-analytics-opt-out', '', array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, false );
		wp_add_inline_script( 'google-analytics-opt-out', $code );
```

Since we are refactoring I'd also rename this function from `header` to something related to opt out to describe it's purpose. It doesn't really output all the header code anyways.

includes/class-wc-google-analytics.php
```diff
@@ -65,21 +69,22 @@ public function __construct() {
...
-		add_action( 'woocommerce_update_options_integration_google_analytics', array( $this, 'process_admin_options') );
-		add_action( 'woocommerce_update_options_integration_google_analytics', array( $this, 'show_options_info') );
-		add_action( 'admin_enqueue_scripts', array( $this, 'load_admin_assets') );
+		add_action( 'woocommerce_update_options_integration_google_analytics', array( $this, 'process_admin_options' ) );
+		add_action( 'woocommerce_update_options_integration_google_analytics', array( $this, 'show_options_info' ) );
+		add_action( 'admin_enqueue_scripts', array( $this, 'load_admin_assets' ) );
 
 		// Tracking code
-		add_action( 'wp_head', array( $this, 'tracking_code_display' ), 999999 );
+		add_action( 'wp_enqueue_scripts', array( $this, 'tracking_code_display' ), 9 );
```
mikkamp: Can we rename this to something like `enqueue_tracking_code` since it doesn't actually do any displaying anymore.

includes/class-wc-google-analytics.php
```diff
@@ -705,4 +698,22 @@ public function add_wc_setup_task() {
...
+	 *
+	 * @see https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/186
+	 *
+	 * @param string $tag HTML for the script tag.
+	 * @param string $handle Handle of script.
+	 * @param string $src Src of script.
+	 *
+	 * @return string
+	 */
+	public function async_script_loader_tag( $tag, $handle, $src ) {
```
mikkamp: It makes sense to prevent the header script from being blocking by adding async, but I'm just curious if that has any side effects.

> `async` : A script that will be downloaded in parallel to parsing the page, and executed as soon as it is available.

Is there a scenario where the code to trigger the tracking will be executed earlier (if the script takes longer to download)? The rest of the code is loaded through inline script where `async` doesn't seem like it would make sense (or have any effect).

I think we also might need a few more checks to make sure the script doesn't have an empty src or async has already been added, because I'm seeing this as the snippet (notice the double `async` for the gtag script):
```
<script async async src='https://www.googletagmanager.com/gtag/js?id=G-NGBB258&#038;ver=1.6.2' id='google-tag-manager-js'></script>
<script async id='google-tag-manager-js-after'>
    gtag('js', new Date());
    ...
```

mikkamp (commented):
Thanks for all the changes. The script is now no longer loaded twice and the loading has become a little more flexible making it easier to adjust. I've added some comments as I went through the code.

Since it's a bit of a major refactor I would make sure we bump at least the minor version in the next release.

includes/class-wc-google-analytics.php
mikkamp: Seems like it's recommended by Google to add async: https://developers.google.com/tag-platform/gtagjs/install
But only on the first snippet. So maybe we need:

```php
		return str_replace( '<script src', '<script async src', $tag );
```

Also not sure about including the plugin version, that could trigger Google to provide the wrong version, maybe we should suppress that in `wp_register_script`.

mikkamp: Just adding a link to what the function looks like now in the blocks implementation: https://github.com/woocommerce/woocommerce-blocks/blob/v9.8.1/src/Domain/Services/GoogleAnalytics.php#L106-L115

The `did_action` condition is not important for us to add since that's refactored now.

includes/class-wc-google-analytics-info-banner.php
puntope: Yes, I forgot to mention that in the PR. It's removed because is not being used. I updated the PR description

includes/class-wc-abstract-google-analytics-js.php
puntope: Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/302924fe1042885e81d23730afe11bcbab1eca1b

includes/class-wc-google-analytics-js.php
puntope: Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/302924fe1042885e81d23730afe11bcbab1eca1b

includes/class-wc-google-analytics-js.php
puntope: You're right. This was done by PHPCBF automatically. I didn't do a deep refactor on those. 

I remove the empty quotes here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/27fd71d12b0b7e624a722f7c23d10d71f70cac88

includes/class-wc-google-analytics.php
puntope: Changed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/ec3401a689054677d2511bf4193ef7b16f658b3b

includes/class-wc-google-analytics.php
puntope: Changed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/ec3401a689054677d2511bf4193ef7b16f658b3b

includes/class-wc-abstract-google-analytics-js.php
puntope: I apply this he name change here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/4b97db4f5d7c89529374da46947109083b6907fb ... however, if I run it via register/add_inline script. It stops working and the script is not being loaded 🤔

includes/class-wc-google-analytics.php
puntope: Thanks for caching this. Added the fix here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/fd7700484b85aa607cd1987574060775ed29cd18

includes/class-wc-abstract-google-analytics-js.php
puntope: In this case is not important to have this one in the footer. Since it's a function that should be called by something... 

`<a href="javascript:gaOptout()">Click here to opt-out of Google Analytics</a>`

But seems is not being used... maybe it was in the past? 🤔

puntope:
Howdy @mikkamp I added all your suggestions except https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260#discussion_r1137780286

I think this is ready for another round

includes/class-wc-abstract-google-analytics-js.php
mikkamp: That's because it also needs a `wp_enqueue_script` to force it to appear. However my suggestion was to add it as a dependency for the `google_tag_manager` script (right now we pass an empty array of dependencies. Making it a dependency forces it to be included before (which is where it used to be).

Ex:
```php
wp_register_script(
	'google-tag-manager',
	'https://www.googletagmanager.com/gtag/js?id=' . esc_js( $gtag_id ),
	array( 'google-analytics-opt-out' ),
	WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION,
	false
);
```

For the `gaOptout()` function it's ok to have it in the footer, but what that does is set a cookie, which is checked on any subsequent pages to disable tracking. If we check the cookie and disable the tracking in the footer then that's too late.

This is the part that must run before the tracking code:
```js
if ( document.cookie.indexOf( disableStr + '=true' ) > -1 ) {
	window[disableStr] = true;
}
```

includes/class-wc-google-gtag-js.php
```diff
@@ -304,12 +306,11 @@ function ' . self::tracker_var() . '(){dataLayer.push(arguments);}
 		} );
 
 		$track_404_enabled
-		</script>
 		";
 
-		$gtag_snippet = apply_filters( 'woocommerce_gtag_snippet' , $gtag_snippet );
+		wp_register_script( 'google-tag-manager', 'https://www.googletagmanager.com/gtag/js?id=' . esc_js( $gtag_id ), array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, false );
+		wp_add_inline_script( 'google-tag-manager', apply_filters( 'woocommerce_gtag_snippet', $gtag_snippet ) );
```
mikkamp, line 312: We also want to add a enqueue call here:

```php
wp_enqueue_script( 'google-tag-manager' );
```

The way this code is now it relies on the enqueue call from blocks `wp_enqueue_script( 'wc-blocks-google-analytics' );` because `google-tag-manager` is a dependent script. It's pretty hard to run a site without blocks but if there is ever any condition where blocks have been disabled, we still want our script to load.

includes/class-wc-google-analytics-js.php
```diff
@@ -256,41 +260,39 @@ protected static function load_analytics_universal( $logged_in ) {
...
 		$ga_snippet_create       = apply_filters( 'woocommerce_ga_snippet_create', $ga_snippet_create, $ga_id );
 		$ga_snippet_developer_id = apply_filters( 'woocommerce_ga_snippet_developer_id', $ga_snippet_developer_id );
 		$ga_snippet_require      = apply_filters( 'woocommerce_ga_snippet_require', $ga_snippet_require );
 
-		$code = "<script type='text/javascript'>" . $ga_snippet_head . $ga_snippet_create . $ga_snippet_developer_id . $ga_snippet_require . "</script>";
-		$code = apply_filters( 'woocommerce_ga_snippet_output', $code );
+		$code = $ga_snippet_head . $ga_snippet_create . $ga_snippet_developer_id . $ga_snippet_require;
 
-		return $code;
+		wc_enqueue_js( apply_filters( 'woocommerce_ga_snippet_output', $code ) );
```
mikkamp: This changes the UA to the footer as well. Technically they have till June to upgrade to a GTag implementation, but we shouldn't be breaking any older sites. I'd prefer to revert these commits and leave these snippets as is. Same applies for the legacy Standard Tracking (function `load_analytics_classic`)

Blocks doesn't have compatibility for UA anyway so no need to resolve the conflict here.

mikkamp (commented):
Thanks for the changes, I left some additional comments as I went through the code.

Although thinking about it again, I think it's a safer bet to only enqueue when they are using the GTag implementation (that's the only one Blocks has compatibility for). Otherwise we risk changing too many old sites where snippets are moved from header to footer. We only really need to provide compatibility for those older sites till end of June, covered in #215

puntope:
Howdy @mikkamp 

I changed the code an now it should works as expected:

GA Opt-out -> Loaded in the header first thing. 
GTAG -> Loaded in Header async -> Avoids duplicates with Blocks 
Analytics Classic -> Loaded in header -> Functions inn the footer
Universal analytics -> Loaded in the header -> Functions in the footer

This is ready for another round.

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -57,45 +58,46 @@ public static function header() {
 			function gaOptout() {
 				document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
 				window[disableStr] = true;
-			}
-		</script>";
+			}";
+
+		wp_register_script( 'google-analytics-opt-out', '', array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, false );
+		wp_add_inline_script( 'google-analytics-opt-out', $code );
+		wp_enqueue_script( 'google-analytics-opt-out' );
```
mikkamp, line 65: Not strictly necessary to enqueue here. I would have left it out as if there are any conditional checks in `load_analytics` which prevent the snippet from enqueueing then it will have already enqueued the opt out snippet. There aren't any of those checks so fine to leave it either way.

includes/class-wc-google-gtag-js.php
```diff
@@ -304,12 +306,11 @@ function ' . self::tracker_var() . '(){dataLayer.push(arguments);}
 		} );
 
 		$track_404_enabled
-		</script>
 		";
 
-		$gtag_snippet = apply_filters( 'woocommerce_gtag_snippet' , $gtag_snippet );
-
-		return $gtag_snippet;
+		wp_register_script( 'google-tag-manager', 'https://www.googletagmanager.com/gtag/js?id=' . esc_js( $gtag_id ), array( 'google-analytics-opt-out' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, false );
```
mikkamp: I would set an empty version here, because we want to get the same script loaded from Google regardless of what our plugin version is at. So we don't want a version string appended to the URL.

The URL is: `https://www.googletagmanager.com/gtag/js?id=G-12345&#038;ver=1.6.2`
Should be just: `https://www.googletagmanager.com/gtag/js?id=G-12345`

woocommerce-google-analytics-integration.php
```diff
@@ -14,6 +14,11 @@
  * Domain Path: languages/
  */
 
+// phpcs:disable Squiz.Classes.ClassDeclaration.SpaceBeforeKeyword
+// phpcs:disable Squiz.Classes.ClassDeclaration.SpaceBeforeCloseBrace
+// phpcs:disable Squiz.Classes.ClassFileName.NoMatch
+// phpcs:disable Squiz.Classes.ValidClassName.NotCamelCaps
```
mikkamp: Can we also move these to the global config?

mikkamp (approved):
Thanks for the additional changes, that's working as expected now. I'd mainly still change the version string in the URL but the others are just minor comments, so I'll go ahead and approve the PR.

woocommerce-google-analytics-integration.php
puntope: Done in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/9e29d0a2216e192fc93f01fdbc16a0763675cd36

includes/class-wc-google-gtag-js.php
puntope: Done in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/260/commits/9e29d0a2216e192fc93f01fdbc16a0763675cd36

Issue #227, opened 2022-09-15
Move GA4 implementation for blocks within the extension


mikkamp:
## User story
The WooCommerce Blocks extension has a separate implementation for supporting tracking events from Blocks, see: https://github.com/woocommerce/woocommerce-blocks/pull/3953
This code is also present within WooCommerce core as part of the blocks package.

Now that this has matured, we shouldn't have to maintain the code for this in a separate location. So it's best to move this code within the Google Analytics Integration extension.

### Potential blockers/points to take into consideration

- Original blocking issue: #186 
- Implementation is in TypeScript whereas this extension is still using jQuery
- Triggers are still tied to an [experimental hook](https://github.com/woocommerce/woocommerce-blocks/blob/v8.5.0/assets/js/extensions/google-analytics/constants.ts#L2), this can be promoted within WC Blocks if needed.

### Describe the solution you'd like 
For all the Google Analytics tracking code to be maintained in the same extension

## Technical
A good example to get references from is how the hooks are used with GLA to add additional tracking: https://github.com/woocommerce/google-listings-and-ads/blob/2.1.2/js/src/gtag-events/index.js

PR #261, opened 2023-03-23
Implement JS Build (ES6) and JS Lint

puntope:

Closes part of #227  .

This PR implements the next major changes.

- Add wp scripts and a basic webpack config for compiling purposes.
- source files are now in `assets/js/src` and the build in  `assets/js/build`
- Exclude source files `assets/js/src` to be included in archived zip file 
- Add js lint support as well as a github workflow for checking so on each PR
- Copy the Blocks actions present in https://github.com/woocommerce/woocommerce-blocks/pull/3953

### Checks:

1. Run `npm i`
2. Run `npm run start`
3. Check the extension loads the `ga-integration.js` file `ga-admin-settings.js` file and they work correctly without errors.
4. Go to one of the files in `assets/js/src` and add a random console.log
5. See how the re-build process runs 
6. Refresh the page and see the console log running
7. Check the es-lint workflow is working fine
8. Try to make some bad changes like undefined constants or bad spacing and see the js linter working.
9. run `npm run archive`
10. See the zip file doesn't contain anymore `asset/js/src` but yes `asset/js/build`
11. Verify the config files 

### Additional details:

⚠️  JS Unit Tests support are not included in this PR
⚠️  A solution for dealing with duplicate actions with blocks will be done in the future PR
⚠️  To discuss if we want to add TypeScript on this project. Seems like a good moment and extension to do so.

### Changelog entry

> Dev - Implement JS Build (ES6) and JS Lint


assets/js/src/ga-integration.js
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line camelcase
+const google_analytics_integration_product_data = [];
```
eason9487: This variable needs to be globally accessible. In addition, it would be a good chance to change to use an object type.


https://user-images.githubusercontent.com/17420811/227476186-2fb1671e-c95a-4227-a13e-0c058e921b7f.mp4

.eslintrc.js
```diff
@@ -0,0 +1,35 @@
...
+	],
+	settings: {
+		jsdoc: {
+			mode: 'typescript',
+		},
+		'import/core-modules': [ 'webpack' ],
+		'import/resolver': { webpack: webpackResolver },
+	},
+	globals: {
+		getComputedStyle: 'readonly',
```
eason9487: `getComputedStyle` seems not to be used in this repo. Instead of removing it, replacing it with `jQuery` could omit the needs of `// eslint-disable-next-line no-undef` in the uses of jQuery.

includes/class-wc-google-analytics.php
```diff
@@ -377,7 +377,7 @@ public function load_admin_assets() {
 			return;
 		}
 
-		wp_enqueue_script( 'wc-google-analytics-admin-enhanced-settings', plugins_url( '/assets/js/admin-ga-settings.js', dirname( __FILE__ ) ), array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
+		wp_enqueue_script( 'wc-google-analytics-admin-enhanced-settings', plugins_url( '/assets/js/build/admin-ga-settings.js', dirname( __FILE__ ) ), array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
```
eason9487: There is an original issue but maybe consider fixing it by the way.

The admin-ga-settings.js script depends on `jQuery` so it should have 'jquery' handle in the `$deps` array.

includes/class-wc-google-gtag-js.php
```diff
@@ -61,7 +61,8 @@ public function setup_frontend_scripts() {
 	 * Register front end JavaScript
 	 */
 	public function register_scripts() {
-		wp_register_script( $this->script_handle, plugins_url( 'assets/js/ga-integration.js', dirname( __FILE__ ) ), array( 'jquery' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
+		wp_register_script( $this->script_handle . '-variable-product-tracking', plugins_url( 'assets/js/build/ga-integration.js', dirname( __FILE__ ) ), array( 'jquery' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
+		wp_enqueue_script( $this->script_handle . '-tracking', plugins_url( 'assets/js/build/actions.js', dirname( __FILE__ ) ), array(), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
```
eason9487: The actions.js script depends on `@wordpress/hooks` and `@wordpress/i18n` packages
so it should have their handles, `'wp-hooks'` and `'wp-i18n'` in the `$deps` array.

A better way would be similar to [GLA's approach](https://github.com/woocommerce/google-listings-and-ads/blob/2.4.1/src/Assets/ScriptWithBuiltDependenciesAsset.php#L49-L69) which gets the `dependencies` array from the auto-generated actions.asset.php file.

.eslintrc.js
```diff
@@ -0,0 +1,35 @@
...
+	extends: [
+		'plugin:@wordpress/eslint-plugin/recommended',
+		'plugin:import/recommended',
+	],
+	settings: {
+		jsdoc: {
+			mode: 'typescript',
+		},
+		'import/core-modules': [ 'webpack' ],
+		'import/resolver': { webpack: webpackResolver },
```
eason9487: The `eslint-import-resolver-webpack` package and this setting are not required if there is no certain alias in webpack config that needs to be resolved in eslint.

Or, maybe consider adding an alias to webpack config like:
```js
resolve: {
	...defaultConfig.resolve,
	alias: {
		'~': path.join( __dirname, 'assets/js/src' ),
	},
},
```

So that the module imports can use it like:
```js
import {
	trackEvent,
	getProductFieldObject,
	getProductImpressionObject,
} from '~/utils';
```

.eslintrc.js
```diff
@@ -0,0 +1,35 @@
...
+	},
+	globals: {
+		getComputedStyle: 'readonly',
+	},
+	rules: {
+		'@wordpress/no-unsafe-wp-apis': 1,
```
eason9487: This rule could be omitted as there is no such linting error.

.gitattributes
eason9487, line 1: .eslintrc.js was missed in setting `export-ignore` and was included in the build bundle.

package.json
```diff
@@ -8,9 +8,17 @@
...
     "url": "git://github.com/woocommerce/woocommerce-google-analytics-integration.git"
   },
+  "dependencies": {
+    "@wordpress/hooks": "^3.22.0",
+    "@wordpress/i18n": "3.15.0"
+  },
   "devDependencies": {
     "node-wp-i18n": "~1.2.3",
-    "uglify-js": "^3.6.0"
+    "uglify-js": "^3.6.0",
```
eason9487: Since the `uglify-js` package is no longer in use, it could be removed.

includes/class-wc-google-gtag-js.php
```diff
@@ -61,7 +61,8 @@ public function setup_frontend_scripts() {
 	 * Register front end JavaScript
 	 */
 	public function register_scripts() {
-		wp_register_script( $this->script_handle, plugins_url( 'assets/js/ga-integration.js', dirname( __FILE__ ) ), array( 'jquery' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
+		wp_register_script( $this->script_handle . '-variable-product-tracking', plugins_url( 'assets/js/build/ga-integration.js', dirname( __FILE__ ) ), array( 'jquery' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
```
eason9487: The `$handle` of the `wp_register_script` and `wp_enqueue_script` functions should be paired. So the `$handle` of `wp_enqueue_script` call at the following reference should be changed accordingly.

https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/23d9af46f00233b70983ad96c67054c18492663e/includes/class-wc-google-gtag-js.php#L56

assets/js/src/constants.js
```diff
@@ -0,0 +1,2 @@
+export const NAMESPACE = 'woocommerce-google-analytics';
```
eason9487: Wonder if it would be better to use the same string as this plugin's slug for this namespace? Or, for somewhat reason, it needs to keep the same namespace after migrating it from WC Blocks?

assets/js/src/actions.js
```diff
@@ -0,0 +1,62 @@
...
+		event_category: 'engagement',
+		event_label: __( 'Viewing products', 'woo-gutenberg-products-block' ),
+		items: products.map( ( product, index ) => ( {
+			...getProductImpressionObject( product, listName ),
+			list_position: index + 1,
+		} ) ),
+	} );
+};
+
+const trackAddToCart = ( product, quantity = 1 ) => {
```
eason9487: The destructuring assignment was missed.

```suggestion
const trackAddToCart = ( { product, quantity = 1 } ) => {
```

assets/js/src/actions.js
```diff
@@ -0,0 +1,62 @@
...
+import { NAMESPACE, ACTION_PREFIX } from './constants';
+import {
+	trackEvent,
+	getProductFieldObject,
+	getProductImpressionObject,
+} from './utils';
+
+const trackListProducts = ( {
+	products,
+	listName = __( 'Product List', 'woo-gutenberg-products-block' ),
```
eason9487: Should it use `'woocommerce-google-analytics-integration'` as the domain string?

assets/js/src/actions.js
```diff
@@ -0,0 +1,62 @@
...
+addAction(
+	`${ ACTION_PREFIX }-set-cart-item-quantity`,
+	NAMESPACE,
+	trackChangeCartItemQuantity
+);
+addAction(
+	`${ ACTION_PREFIX }-remove-cart-item`,
+	NAMESPACE,
+	trackRemoveCartItem
+);
```
eason9487: The hook names have been changed.

```suggestion
addAction( `${ ACTION_PREFIX }-product-list-render`, NAMESPACE, trackListProducts );
addAction( `${ ACTION_PREFIX }-cart-add-item`, NAMESPACE, trackAddToCart );
addAction(
	`${ ACTION_PREFIX }-cart-set-item-quantity`,
	NAMESPACE,
	trackChangeCartItemQuantity
);
addAction(
	`${ ACTION_PREFIX }-cart-remove-item`,
	NAMESPACE,
	trackRemoveCartItem
);
```

assets/js/src/utils.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Returns the product ID by checking if the product has a SKU, if not, it returns '#' concatenated with the product ID.
+ *
+ * @param {Object} product - The product object
+ *
+ * @return {string} - The product ID
+ */
+const getProductId = ( product ) => {
+	return product?.sku ?? '#' + product.id;
```
eason9487: Is `product` possible a `null` or `undefined`? If yes, `product.id` will cause an error; if no, the optional chaining operator is redundant.

In addition, even if changing to `product.sku ?? '#' + product.id` is not exactly equivalent to `product.sku ? product.sku : '#' + product.id` when the `product.sku` is an empty string.

eason9487 (changes requested):
Thanks for the work.

Unfortunately, I was hardly able to complete the test.

Found some critical issues:
- The `assets/js/src/ga-integration.js` script is not loaded to the page. And even if after fixing the loading issue, the "google_analytics_integration_product_data is not defined" errors will occur.
- All the event tracking in `assets/js/src/actions.js` are not fired since the mismatched hook names. Further, even if fix the hook names, "Cannot read properties of undefined (reading 'price')" errors will occur when calling the `trackAddToCart` function.
- [The event tracking in WC Blocks](https://github.com/woocommerce/woocommerce-blocks/blob/v9.8.2/assets/js/extensions/google-analytics/index.ts) has changed quite a lot, for example, the number of events has increased from 4 to 16. Does it need to sync the subsequent changes?

Other issues that are also worth noting:
- Some project configurations could be omitted or enhanced.
- Two scripts' `$deps` array is not set properly.

.eslintrc.js
puntope: Thanks. Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/f61ea20641f696e4320787f05a22dca4af6fa1f0

package.json
puntope: Thanks, fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/da1284f374ae36a5c394622c764c8ab460f4a7c6 I also added prettier in this commit

.gitattributes
puntope: Thanks, fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/67d7e8986ad4886b41a63a95e77c3eb066134b20

.eslintrc.js
puntope: Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/ad25ad98f5cea683e9f8eef24f5126a25360df1f thanks

.eslintrc.js
puntope: Fixed here thanks https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/ad25ad98f5cea683e9f8eef24f5126a25360df1f

includes/class-wc-google-gtag-js.php
puntope: Good idea. Added here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/f3e783bb9d1136c99eca96c2aa412868fe027c18

includes/class-wc-google-analytics.php
puntope: Should be resolved here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/f3e783bb9d1136c99eca96c2aa412868fe027c18

includes/class-wc-google-gtag-js.php
puntope: Right. Thanks for this catch. Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/f3e783bb9d1136c99eca96c2aa412868fe027c18

assets/js/src/ga-integration.js
puntope: Interesting. I added `window.` in the creation of the variable and also use object instead of array. 

https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/be572242209a248a42db9fd8b5b1b66c5d443f2d

assets/js/src/actions.js
puntope: Thanks for catching this one. It's fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/be572242209a248a42db9fd8b5b1b66c5d443f2d

assets/js/src/constants.js
puntope: I changed it here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/d30bce6b2053cc6b2c4207f4444ed448e7b90b71

assets/js/src/actions.js
puntope: Thanks! Added here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/890cee55cbab9d497427ce0422788ebf20f2cc86

assets/js/src/utils.js
puntope: Right. Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/38214eeb99077de2a7d065b4864682123b43fc00

puntope:
Howdy @eason9487 first of all, thanks for the amazing review as always. 

I patched all of your findings. Still need to test deeper to see everything is working find. But I think at least we can  have another round to check most of the issues :) 

> [The event tracking in WC Blocks](https://github.com/woocommerce/woocommerce-blocks/blob/v9.8.2/assets/js/extensions/google-analytics/index.ts) has changed quite a lot, for example, the number of events has increased from 4 to 16. Does it need to sync the subsequent changes?

Oh, I see. Maybe we can keep this one like it is (with the reduced amount) without merge and make those in a future PR if thats ok.

package.json
```diff
@@ -8,9 +8,17 @@
...
   },
+  "dependencies": {
+    "@wordpress/hooks": "^3.22.0",
+    "@wordpress/i18n": "3.15.0"
+  },
   "devDependencies": {
+    "@wordpress/eslint-plugin": "^8.0.0",
+    "@wordpress/scripts": "^24.6.0",
+    "eslint": "^7.17.0",
+    "eslint-import-resolver-webpack": "^0.13.1",
```
eason9487: The `eslint-import-resolver-webpack` package is not required after https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/ad25ad98f5cea683e9f8eef24f5126a25360df1f.

assets/js/src/actions.js
eason9487: As a record, I think this has been fixed by 108a8563b593bccc71f541c1dfdc097412724ddb.

package.json
```diff
@@ -19,16 +27,17 @@
...
-    "build": "npm run uglify && npm run makepot && npm run archive",
-    "build:dev": "npm run uglify && npm run makepot",
+    "dev": "NODE_ENV=development wp-scripts build",
+    "start": "wp-scripts start",
+    "build": "NODE_ENV=production wp-scripts build && npm run makepot &&  npm run archive",
+    "makepot": "wpi18n makepot --domain-path languages --pot-file $npm_package_name.pot --type plugin --main-file $npm_package_name.php --exclude node_modules,tests,docs,assets/js/src",
+    "lint:js": "wp-scripts lint-js",
+    "lint:php": "vendor/bin/phpcs",
+    "lint:php:diff": "./bin/phpcs-diff.sh",
+    "prearchive": "rm -rf vendor && composer install --no-dev && composer dump-autoload -o",
```
eason9487: This script causes the `vendor` directory to be included in the build bundle.

includes/class-wc-google-analytics.php
eason9487: `Plugin::get_instance()->get_js_asset_dependencies( 'admin-ga-settings' )` returns an empty array. The 'jquery' handle is still missing.

includes/class-wc-google-gtag-js.php
```diff
@@ -53,15 +55,28 @@ public function setup_frontend_scripts() {
...
 
 	/**
 	 * Register front end JavaScript
 	 */
 	public function register_scripts() {
-		wp_register_script( $this->script_handle, plugins_url( 'assets/js/ga-integration.js', dirname( __FILE__ ) ), array( 'jquery' ), WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, true );
+		wp_register_script(
+			$this->script_handle . '-ga-integration',
+			Plugin::get_instance()->get_js_asset_url( 'ga-integration.js' ),
+			Plugin::get_instance()->get_js_asset_dependencies( 'ga-integration' ),
```
eason9487: The 'jquery' handle is missing after changing to use `get_js_asset_dependencies`.

woocommerce-google-analytics-integration.php
```diff
@@ -204,6 +204,84 @@ public function maybe_show_ga_pro_notices() {
...
+		 * @param string $end End of the URL.
+		 *
+		 * @return string
+		 */
+		public function url( $end = '' ) {
+			return untrailingslashit( plugin_dir_url( plugin_basename( __FILE__ ) ) ) . $end;
+		}
+
+		/**
+		 * Get the URL to something in the plugin admin assets dir.
```
eason9487: Maybe adjusting it to indicate it's the URL to the build directory of js assets would be nicer.

woocommerce-google-analytics-integration.php
```diff
@@ -204,6 +204,84 @@ public function maybe_show_ga_pro_notices() {
...
+		 * @param string $end End of the URL.
+		 *
+		 * @return string
+		 */
+		public function get_js_asset_url( $end = '' ) {
+			return $this->url( '/assets/js/build/' . $end );
+		}
+
+		/**
+		 * Get the path to something in the plugin admin dir.
```
eason9487: Similar to the above. Would be better to indicate it's the path to the build directory of js assets?

woocommerce-google-analytics-integration.php
```diff
@@ -204,6 +204,84 @@ public function maybe_show_ga_pro_notices() {
...
+
+		/**
+		 * Gets the asset.php generated file for an asset name.
+		 *
+		 * @param string $asset_name The name of the asset to get the file from.
+		 * @return array The asset file. Or an empty array if the file doesn't exist.
+		 */
+		public function get_js_asset_file( $asset_name ) {
+			try {
+				return require $this->get_js_asset_path( $asset_name . '.asset.php' );
```
eason9487, line 258: Only the handle of DEWPed packages  imported via the `import` declarations will be added to the `dependencies` in the *.asset.php file. This won't resolve the `$deps` for JS files that use `jQuery` directly.

eason9487 (commented):
> > [The event tracking in WC Blocks](https://github.com/woocommerce/woocommerce-blocks/blob/v9.8.2/assets/js/extensions/google-analytics/index.ts) has changed quite a lot, for example, the number of events has increased from 4 to 16. Does it need to sync the subsequent changes?
> 
> Oh, I see. Maybe we can keep this one like it is (with the reduced amount) without merge and make those in a future PR if thats ok.

Keeping it as it's would be OK but I guess it would need a feature flag to skip the `wp_enqueue_script` call by default for the actions.js file, or even remove the whole `wp_enqueue_script` call for now. So that it won't lead to duplicate event tracking with WC Blocks.

woocommerce-google-analytics-integration.php
puntope: I added a workaround here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/450963ebb6e324df566f5c56c7bf8eab298ef00d

includes/class-wc-google-gtag-js.php
puntope: I added a workaround here https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/450963ebb6e324df566f5c56c7bf8eab298ef00d

package.json
puntope: Fixed here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/3b5ea432cb3e9bc07ad2c8a19c39b69eefc07310

woocommerce-google-analytics-integration.php
puntope: Tweaked here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/f2689e15d5ccc1a5637a1f7d026346722c5dc4a1

woocommerce-google-analytics-integration.php
puntope: Tweaked here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/f2689e15d5ccc1a5637a1f7d026346722c5dc4a1

package.json
puntope: Can you elaborate this? I don't understand the issue

puntope:
> > > [The event tracking in WC Blocks](https://github.com/woocommerce/woocommerce-blocks/blob/v9.8.2/assets/js/extensions/google-analytics/index.ts) has changed quite a lot, for example, the number of events has increased from 4 to 16. Does it need to sync the subsequent changes?
> > 
> > 
> > Oh, I see. Maybe we can keep this one like it is (with the reduced amount) without merge and make those in a future PR if thats ok.
> 
> Keeping it as it's would be OK but I guess it would need a feature flag to skip the `wp_enqueue_script` call by default for the actions.js file, or even remove the whole `wp_enqueue_script` call for now. So that it won't lead to duplicate event tracking with WC Blocks.

I created the feature `feature/analytics-wc-blocks` so we can safely merge everything there instead of trunk.

puntope:
Howdy @eason, I added new patches and some question in regards your comment here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261#discussion_r1158298241

package.json
eason9487: After running `npm run build`, the vendor directory is bundled into the woocommerce-google-analytics-integration.zip file. The `vendor` directory is not included in the past releases, so I think this is probably not the expected result.

P.S. `composer archive` has different results in different versions. I'm using version 2.5.2

image

puntope: Ahhh ok, I see now that composer has not required libraries so it's no need of it of being executed. I removed it in here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/261/commits/33e19c8e5e77ab23d4b78f69dd9c3f58495e1397

eason9487 (approved):
Thanks for the further adjustment. LGTM.

PR #280, opened 2023-06-26
Set Node/npm Engines

puntope:

- Set the strict-engines directive in `.npmrc `
- Set the engines in `package.json`

1.   Run `nvm use 12` and see the npm compilation builds
2.  Run `nvm use 14` and see the npm compilation builds
3.  Run `nvm use 10` (or any other version,< 12)  and see the npm compilation fails

### Changelog entry

> Dev - Set engines for the repository

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for adding this. I tested with Node 10 to 18, only Node 10 had build failed.

PR #287, opened 2023-08-02
Fix script tag replacer

### LINKED ISSUE #279: Change to use wp_enqueue has made gtag.js load synchronously
### Describe the bug:
The commit made in https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/730495a09bbb22798f897a64907ceeccf278dcb5#diff-75fada6b677c39f17a7de173a315db4d8911e7469f5a70dce4822a6aaa849514 has switched to loading gtag using wp_queue_script. Unfortunately this has meant gtag lost its async tag and is loading render blocking. Scores on Core Web Vitals show. 

### Steps to reproduce:
1. Upgrade to a version of the plugin post this commit
2. View source
3. gtag script has no async tag

### Expected behavior:
Tag should load async to improve Woocommerce performance

### Actual behavior:
Change regressed performance due to synchronous JS

### Additional details:
image

puntope:

Closes #279 

### Checks:

1. Dont check out this PR and create this filter
```php
add_filter( 'script_loader_tag', array( $this, 'change_script_tags' ), 1, 3 );
function change_script_tags( $tag, $handle, $src ) {
		if ( in_array( $handle, self::ASYNC_SCRIPT_HANDLES, true ) ) {
			$tag = '<script type="text/javascript" src="' . $src . '" id="' . $handle . '-js"></script>';
		}
		return $tag;
	}
```
2. Go to the Shop and search for `google-tag-manager` script in the source
3. See "async" attribute not being added
4. Checkout this PR
5. Refresh
6.  See "async" attribute being added

### Changelog entry

> Fix - Add async attribute in `google-tag-manager` script.

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -723,7 +723,7 @@ class WC_Google_Analytics extends WC_Integration {
 		}
 
 		// Check if the script has the async attribute already. If so, don't add it again.
-		$has_async_tag = preg_match('/^.*async.*$/', $tag);
+		$has_async_tag = preg_match( '/^.*async.*$/', $tag );
 		if ( ! empty( $has_async_tag ) ) {
 			return $tag;
 		}
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -722,6 +722,13 @@ class WC_Google_Analytics extends WC_Integration {
 			return $tag;
 		}
 
+		// Check if the script has the async attribute already. If so, don't add it again.
+		$has_async_tag = preg_match('/^.*async.*$/', $tag);
+		if ( ! empty( $has_async_tag ) ) {
+			return $tag;
+		}
+
+		// Add the async attribute
 		return str_replace( ' src', ' async src', $tag );
 	}
 }
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -722,6 +722,6 @@ class WC_Google_Analytics extends WC_Integration {
 			return $tag;
 		}
 
-		return str_replace( '<script src', '<script async src', $tag );
+		return str_replace( ' src', ' async src', $tag );
 	}
 }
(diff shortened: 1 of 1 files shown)


eason9487 (dismissed):
Thanks for the fix.

This is not a blocker, but it would be a bit better to avoid having two `async` attributes, even though it won't cause any real problems. (One of the `async` attributes seems to come from WC-Blocks)

image

puntope:
Howdy @eason9487 thanks for that catch. Might you test again with this change https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/287/commits/f5712943354124b17204cf11f2b5e72bf3f9d1b6 ?

Thanks

includes/class-wc-google-analytics.php
```diff
@@ -722,6 +722,13 @@ public function async_script_loader_tags( $tag, $handle, $src ) {
 			return $tag;
 		}
 
-		return str_replace( '<script src', '<script async src', $tag );
+		// Check if the script has the async attribute already. If so, don't add it again.
+		$has_async_tag = preg_match( '/^.*async.*$/', $tag );
```
eason9487: Two `async` attributes still occur as this regex pattern cannot match correctly. Maybe consider matching it with a pattern like `/\basync\b/`?

puntope:
Thanks @eason9487, for me it seems like working in Firefox and Chrome... but I replaced it with your solution here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/287/commits/e3dd72ce11a4a0bb3b7b04c011f3be2a4be5c7c8 

Can you check one more time?
Thanks

eason9487 (approved):
Thanks for the fix. LGTM.


> for me it seems like working in Firefox and Chrome... [...]

Not sure what makes them different. In my testing env, the `$tag` content includes two `script` tags at a time and is separated by a linebreak char as the following, so the previous regex pattern couldn't match.

```html
<script async src='https://www.googletagmanager.com/gtag/js?id=G-XXXXXXXXXX' id='google-tag-manager-js'></script>
<script id='google-tag-manager-js-after'>
	gtag('js', new Date());
	gtag('config', 'G-XXXXXXXXXX', { 'send_page_view': false });
</script>
```

PR #293, opened 2023-08-21
Add Workflow for generating Hooks Documentation

puntope:

This PR creates an Action for generating PHP Hooks documentation when PHP files change on a new release/branch.

Git Bot will commit a Hooks.md in docs/Hooks.md (empty file was added to this repo)

The workflow is based in the approved solution in https://github.com/woocommerce/google-listings-and-ads/pull/2060

### Changelog entry

> Dev - Add Workflow for generation Hooks documentation

### DIFF
--- a/.github/workflows/php-hook-documentation.yml
+++ b/.github/workflows/php-hook-documentation.yml
@@ -5,16 +5,7 @@ on:
     branches:
       - "release/**"
     paths:
-      - "**.php"
-      - .github/workflows/php-hook-documentation.yml
-  pull_request:
-    types:
-      - opened
-    branches:
-      - "release/**"
-    paths:
-      - "**.php"
-      - .github/workflows/php-hook-documentation.yml
+      - changelog.txt # Run the workflow only after the last commit of Woo Release. When it updates the changelog.
   workflow_dispatch:
 
 concurrency:
@@ -37,7 +28,6 @@ jobs:
         id: generate-hook-docs
         uses: woocommerce/grow/hook-documentation@actions-v1
         with:
-          debug-output: yes
           source-directories: includes/ woocommerce-google-analytics-integration.php
 
       - name: Commit hook documentation
--- a/.github/workflows/php-hook-documentation.yml
+++ b/.github/workflows/php-hook-documentation.yml
@@ -28,7 +28,7 @@ jobs:
         id: generate-hook-docs
         uses: woocommerce/grow/hook-documentation@actions-v1
         with:
-          source-directories: includes/ woocommerce-google-analytics-integration.php
+          source-directories: includes/,woocommerce-google-analytics-integration.php
 
       - name: Commit hook documentation
         shell: bash
--- a/.github/workflows/php-hook-documentation.yml
+++ b/.github/workflows/php-hook-documentation.yml
@@ -1,0 +1,58 @@
+name: PHP Hook Documentation Generator
+
+on:
+  push:
+    branches:
+      - "release/**"
+    paths:
+      - "**.php"
+      - .github/workflows/php-hook-documentation.yml
+  pull_request:
+    types:
+      - opened
+    branches:
+      - "release/**"
+    paths:
+      - "**.php"
+      - .github/workflows/php-hook-documentation.yml
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  HookDocumentation:
+    name: Hook Documentation Generator
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+        with:
+          # Checks out a branch instead of a commit in detached HEAD state
+          ref: ${{ github.head_ref }}
+
+        # This generates the documentation string. The `id` property is used to reference the output in the next step.
+      - name: Generate hook documentation
+        id: generate-hook-docs
+        uses: woocommerce/grow/hook-documentation@actions-v1
+        with:
+          debug-output: yes
+          source-directories: includes/ woocommerce-google-analytics-integration.php
+
+      - name: Commit hook documentation
+        shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/php-hook-documentation.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          # Checks out a branch instead of a commit in detached HEAD state
+          ref: ${{ github.head_ref }}
+
+        # This generates the documentation string. The `id` property is used to reference the output in the next step.
+      - name: Generate hook documentation
+        id: generate-hook-docs
+        uses: woocommerce/grow/hook-documentation@actions-v1
+        with:
+          debug-output: yes
```
eason9487: This parameter seems not to be needed as it's for debug purposes.

.github/workflows/php-hook-documentation.yml
```diff
@@ -0,0 +1,58 @@
...
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+          echo "${{ steps.generate-hook-docs.outputs.hook-docs }}" > docs/Hooks.md
+          git add docs/Hooks.md
+          if git diff --cached --quiet; then
+            echo "*No documentation changes to commit.*" >> $GITHUB_STEP_SUMMARY
+          else
+            echo "*Committing documentation changes.*" >> $GITHUB_STEP_SUMMARY
+            git commit -q -m "Update hooks documentation from ${{ github.head_ref }} branch."
+            git push
```
eason9487, line 47: This git operation may fail because the *woorelease* tool separately pushes two commits in a short period of time:
1. https://github.com/woocommerce/google-listings-and-ads/commit/aa9c94e4888dcd865d9028e2fe90f9c868513590
2. https://github.com/woocommerce/google-listings-and-ads/commit/13f5eee97e2d83d876cd2cda24c025a8be5c08cd

The first one changed a PHP file to trigger the Hook Documentation workflow. The second one only changed non-PHP files so it didn't trigger the same workflow. However, because of the time gap, the `commit` is not on the latest branch head, so the `push` was rejected.

image
Ref: https://github.com/woocommerce/google-listings-and-ads/actions/runs/5934414180/job/16091282802#step:4:862

.github/workflows/php-hook-documentation.yml
```diff
@@ -0,0 +1,58 @@
...
+      - "**.php"
+      - .github/workflows/php-hook-documentation.yml
+  pull_request:
+    types:
+      - opened
+    branches:
+      - "release/**"
+    paths:
+      - "**.php"
+      - .github/workflows/php-hook-documentation.yml
```
eason9487: Under the current release process, this event will not trigger this workflow because GitHub Actions is intentionally designed to prevent workflows from being chain-triggered.

For example, in https://github.com/woocommerce/google-listings-and-ads/pull/2067, the same workflow in that repo was not triggered after `github-actions` bot opened that PR, and then I triggered it manually: https://github.com/woocommerce/google-listings-and-ads/actions/runs/5934182562.

A possible trick is to use the [`workflow_run` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run). Another way is to remove this block and just rely on the flow of *woorelease* tool because it includes a step to bump product version in PHP file(s) so that this workflow will be triggered by `push` event eventually.

eason9487 (commented):
Thanks for adding the new workflow.

As it doesn't work well with the current release process, it might need some adjustments.

puntope:
Hi @eason9487  Thanks for the review and for finding out about this issue. When I saw @github-actions
[Update hooks documentation from branch.](https://github.com/woocommerce/google-listings-and-ads/pull/2067/commits/1d05c13e98b46a9985d16121b122dca7260cd2b8) I took as granted it was triggered automatically.

I updated the trigger for running only when changelog.txt is updated in the push. To be sure it only triggers once in the last commit. However, when I tested it, I realized that Hooks Documentation Generator is not really working for this repo. It seems like a separated issue from the mentioned one.

puntope:
@eason9487 I suggest continuing the conversation about the trigger in https://github.com/woocommerce/automatewoo/pull/1537

As Hook Documentation Generator seems to have EOF problems with this repo and that will require more investigation 🤔

.github/workflows/php-hook-documentation.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          # Checks out a branch instead of a commit in detached HEAD state
+          ref: ${{ github.head_ref }}
+
+        # This generates the documentation string. The `id` property is used to reference the output in the next step.
+      - name: Generate hook documentation
+        id: generate-hook-docs
+        uses: woocommerce/grow/hook-documentation@actions-v1
+        with:
+          source-directories: includes/ woocommerce-google-analytics-integration.php
```
eason9487: > However, when I tested it, I realized that Hooks Documentation Generator is not really working for this repo.

I pushed 5a822245229a55c0f3309ad4be97ff7116983949 to #294 and the document was updated successfully. See: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/294/files/e2621507048da79b2285b45417da91a28c36b2c2..1764a5bdd881e577be99fe52ba9daa2cd7e4b37a

l think this should fix the issue:

```suggestion
          source-directories: includes/,woocommerce-google-analytics-integration.php
```

eason9487 (approved):
Thanks for the fix!

> However, when I tested it, I realized that Hooks Documentation Generator is not really working for this repo.

Just noticed it probably due to a missing comma. It still needs a fixed but the fix is very intuitive and clear. I believe this PR can be approved earlier to minimize the back-and-forth time for code reviews.

> As Hook Documentation Generator seems to have EOF problems with this repo and that will require more investigation

Yes, I guess so.

.github/workflows/php-hook-documentation.yml
puntope: Thanks @eason9487 , still in low mode after the fever so I missed that!

PR #298, opened 2023-09-11
Fix Add To Cart and Impression events when using Blocks

### LINKED ISSUE #297: [Product Blocks] - Add To Cart Tracking Not Working in Shop Product Catalog
### Describe the bug:

When using Product Blocks (using shortcode or a theme like TT3) the add_to_cart event on product catalog doesn't work.

https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/e2b0d75b-986c-4fe1-8410-56a513a69821

### Steps to reproduce:

1. Install non blocks theme like TT1
2. Go to shop and click in Add to cart
3. See the Add to cart event triggered (you can use Google Analytics Debugger or check the network)
4. Install blocks theme like TT3
5. Go to shop and click in Add to cart
6. See the Add to cart event not being triggered 

### Expected behavior:

Add to cart is triggered also in Blocks

### Actual behavior:

Add to cart is not triggered in Blocks

### Additional details:

We did some work in GLA solving a similar issue
https://github.com/woocommerce/google-listings-and-ads/pull/2028

puntope:

Closes #297 

This PR fixes a bug in Products Blocks and Add To cart / Impression tracking.

### Checks:

1. Install Google Analytics Debugger and Checkout this PR
2. Install a theme supporting blocks (like tt3)
3. Go to Shop and check that the impressions are being tracked
4. Click Add To Cart and check that the event is being tracked
5. Install a theme not supporting blocks (like tt1)
6. Repeat steps 3 and 4
7. Create a new Page and add the block "All products"
8. Repeat steps 3 and 4
9. Create a new Page and add the block "Products (beta)"
10. Repeat steps 3 and 4

### Additional details:

One error is happening because the class is not covering `button` tags... only `a` tags
The second problem is because `woocommerce_after_shop_loop_item` is not happening in blocks. But yes  `woocommerce_after_shop_loop_item_title`

### Changelog entry

> Fix - Add To Cart and Impression events when using Blocks

### DIFF
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -226,7 +226,7 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 
 		wc_enqueue_js(
 			"
-			$( '.product.post-" . esc_js( $product->get_id() ) . " a , .product.post-" . esc_js( $product->get_id() ) . " button' ).on('click', function() {
+			$( '.product.post-" . esc_js( $product->get_id() ) . ' a , .product.post-' . esc_js( $product->get_id() ) . " button' ).on('click', function() {
 				if ( true === $(this).hasClass( 'add_to_cart_button' ) ) {
 					$add_to_cart_event_code
 				} else {
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -226,7 +226,7 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 
 		wc_enqueue_js(
 			"
-			$( '.products .post-" . esc_js( $product->get_id() ) . " a' ).on('click', function() {
+			$( '.product.post-" . esc_js( $product->get_id() ) . " a , .product.post-" . esc_js( $product->get_id() ) . " button' ).on('click', function() {
 				if ( true === $(this).hasClass( 'add_to_cart_button' ) ) {
 					$add_to_cart_event_code
 				} else {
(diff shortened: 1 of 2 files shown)


includes/class-wc-google-gtag-js.php
```diff
@@ -226,7 +226,7 @@ public static function listing_click( $product, $position ) {
 
 		wc_enqueue_js(
 			"
-			$( '.products .post-" . esc_js( $product->get_id() ) . " a' ).on('click', function() {
+			$( '.product.post-" . esc_js( $product->get_id() ) . ' a , .product.post-' . esc_js( $product->get_id() ) . " button' ).on('click', function() {
```
martynmjones, line 229: On the `Products (beta)` block page it doesn't look like any of this inline code is included when viewing the page source. Tested with TT1 and TT3.

martynmjones (commented):
Hey @puntope, thanks for your work on this compatibility issue!

I ran into problems with steps 9 and 10 of the testing instructions where neither the impression nor add to cart event is triggered.

It works fine when using the `All products` block but not the `Products (beta)` block 🤔

includes/class-wc-google-gtag-js.php
puntope: Hi @martynmjones  sorry I forgot that Products Beta works differently and It will be needed to add another separated snippet because it doesn't include hooks. 

I think we can for now merge this PR which solves Block Based themes and "All products" blocks and address the PRoducts Beta in a separated PR. This is because 8.1 was released already. 

I created the issue here--- I will go for it tomorrow 
https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/299

WDYT?

martynmjones: Sounds good @puntope. There were no other issues so this PR should be good to merge then 👍

martynmjones (approved):

PR #307, opened 2023-10-02
Enable since tag replacement

### LINKED ISSUE #304: Using since tags without replacements
### Describe the bug:
In the extensions our team manages we use `@since x.x.x` which should be replaced when using WooRelease with the released version. However it seems this is currently not activated for this extension.

When searching for x.x.x in the code there are several occurrences where it hasn't been replaced by the correct version. We should check through the PR's to back fill this information with the correct version and then enable the automatic replacement by WooRelease.

mikkamp:

This PR enables since tag replacements for WooRelease. It also updates any of the previously missed tags to the corresponding version when the PR was released.

Closes #304

1. Confirm there are no more occurrences of the string x.x.x
2. Check the version replace settings are similar to the ones we use in [other repos](https://github.com/woocommerce/google-listings-and-ads/blob/2.5.7/package.json#L127-L133)

### Changelog entry
* Dev - Enable since tag replacement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,12 @@
     "npm": ">=6"
   },
   "config": {
-    "wp_org_slug": "woocommerce-google-analytics-integration"
+    "wp_org_slug": "woocommerce-google-analytics-integration",
+    "version_replace_paths": [
+      "src",
+      "tests",
+      "bin",
+      "woocommerce-google-analytics-integration.php"
+		]
   }
 }
--- a/package.json
+++ b/package.json
@@ -44,9 +44,8 @@
   "config": {
     "wp_org_slug": "woocommerce-google-analytics-integration",
     "version_replace_paths": [
-      "src",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -42,6 +42,12 @@
     "npm": ">=6"
   },
   "config": {
-    "wp_org_slug": "woocommerce-google-analytics-integration"
+    "wp_org_slug": "woocommerce-google-analytics-integration",
+    "version_replace_paths": [
+      "src",
```
puntope: It should be `includes` instead of `src`

`bin` seems not necessary as we are only adding sh files.

mikkamp: Ofcourse, it should be `includes`. Not sure why I didn't replace that.

puntope (approved):

puntope:
✅  LGTM

PR #309, opened 2023-10-03
Add product identifier setting

### LINKED ISSUE #295: Use product_id instead of sku
Is it possible to have a flag so that `get_product_identifier` returns `$product->get_id()` instead of `$product->get_sku()`?

The google merchant centre product feed is setup such that the merchant centre item_id is the woocommerce product id instead of the sku. 

This causes misalignment in google analytics.

Also why is the product id prepended by a hash? Does this get parsed in google analytics and linked to the google merchant centre item correctly?

martynmjones:

Closes #295 

Adds a new setting to `WooCommerce > Settings > Integration > Google Analytics` that allows merchants to specify the structure of the product identifier for Google Analytics.

### Checks:

### Screenshots:

Screenshot 2023-10-04 at 10 05 28

1. On a test site that already has Google Analytics Integration configured, deactivate the extension, checkout `add/295-product-identifier-setting`, and re-activate the extension
2. Go to `WooCommerce > Settings > Integration > Google Analytics` and confirm that `Product Identification` is set to `Product SKU with prefixed (#) ID as fallback`
3. Install the extension from this branch on a new test site that has never had Google Analytics Integration installed and configured
4. Go to `WooCommerce > Settings > Integration > Google Analytics` and confirm that `Product Identification` is set to `Product ID` (_The default for new installs_)
5. Enter a value for `Google Analytics Tracking ID`, check `Enable Enhanced eCommerce`, and `Save changes`
6. Visit a product incognito that has a SKU set with GA Debug enabled
7. Check the console to confirm that for the `view_item` event, the item `id` is the product ID
8. In `wp-admin`, change the setting to `Product SKU with prefixed (#) ID as fallback`
9. Refresh the product page incognito and confirm that the SKU is now used for the `view_item` event
10. Delete the SKU for this product
11. Refresh the product page incognito and confirm that the product ID prefixed with a # is now used for the `view_item` event
12. Add a snippet to the site:
```php
add_filter(
	'woocommerce_ga_product_identifier',
	function( $id, $product ) {
		return $product->get_price() . '-test';
	},
	10,
	2
);
```
13. Refresh the product page incognito and confirm that the value set using the filter is now used for the `view_item` event

### Additional details:

This change does not yet apply to tracking when using WooCommerce Blocks. That will be addressed in a follow-up PR.

### Changelog entry

> Add - Setting to specify the structure of the product identifier

### DIFF
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -31,8 +31,9 @@ abstract class WC_Abstract_Google_Analytics_JS {
 	/**
 	 * Return one of our options
 	 *
-	 * @param  string $option Key/name for the option
-	 * @return string         Value of the option
+	 * @param string $option Key/name for the option
+	 *
+	 * @return string|null Value of the option or null if not found
 	 */
 	protected static function get( $option ) {
 		return self::$options[ $option ] ?? null;
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -35,7 +35,7 @@ abstract class WC_Abstract_Google_Analytics_JS {
 	 * @return string         Value of the option
 	 */
 	protected static function get( $option ) {
-		return self::$options[ $option ];
+		return self::$options[ $option ] ?? null;
 	}
 
 	/**
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -114,11 +114,17 @@ abstract class WC_Abstract_Google_Analytics_JS {
 	 * @return string
 	 */
 	public static function get_product_identifier( $product ) {
-		if ( ! empty( $product->get_sku() ) ) {
-			return esc_js( $product->get_sku() );
-		} else {
-			return esc_js( '#' . ( $product->is_type( 'variation' ) ? $product->get_parent_id() : $product->get_id() ) );
+		$identifier = $product->get_id();
+
+		if ( 'product_sku' === self::get( 'ga_product_identifier' ) ) {
+			if ( ! empty( $product->get_sku() ) ) {
+				$identifier = $product->get_sku();
+			} else {
+				$identifier = '#' . ( $product->is_type( 'variation' ) ? $product->get_parent_id() : $product->get_id() );
+			}
 		}
+
+		return apply_filters( 'woocommerce_ga_product_identifier', $identifier, $product );
 	}
 
 	/**
--- a/woocommerce-google-analytics-integration.php
+++ b/woocommerce-google-analytics-integration.php
@@ -30,6 +30,7 @@ if ( ! class_exists( 'WC_Google_Analytics_Integration' ) ) {
 		__FILE__,
 		function () {
 			WC_Google_Analytics_Integration::get_instance()->maybe_show_ga_pro_notices();
+			WC_Google_Analytics_Integration::get_instance()->maybe_set_defaults();
 		}
 	);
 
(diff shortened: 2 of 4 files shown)


includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -35,7 +35,7 @@ abstract public static function get_instance( $options = array() );
 	 * @return string         Value of the option
 	 */
 	protected static function get( $option ) {
-		return self::$options[ $option ];
+		return self::$options[ $option ] ?? null;
```
puntope, line 39: Maybe we can update the PHPDOC so now this function can return NULL

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -114,11 +114,17 @@ abstract protected function add_transaction_enhanced( $order );
 	 * @return string
 	 */
 	public static function get_product_identifier( $product ) {
-		if ( ! empty( $product->get_sku() ) ) {
-			return esc_js( $product->get_sku() );
-		} else {
-			return esc_js( '#' . ( $product->is_type( 'variation' ) ? $product->get_parent_id() : $product->get_id() ) );
+		$identifier = $product->get_id();
+
+		if ( 'product_sku' === self::get( 'ga_product_identifier' ) ) {
```
puntope, line 120: Maybe add some constant/getter for `product_sku` literal?

woocommerce-google-analytics-integration.php
```diff
@@ -208,6 +209,26 @@ public function maybe_show_ga_pro_notices() {
...
+		 * @return void
+		 */
+		public function maybe_set_defaults() {
+			$settings_key = 'woocommerce_google_analytics_settings';
+
+			if ( false === get_option( $settings_key, false ) ) {
+				update_option(
+					$settings_key,
+					array(
+						'ga_product_identifier' => 'product_id',
```
puntope, line 226: Shouldn't be this `product_sku`? As the default setting?

puntope: I see it's intentional for new install but just wonder why we change from SKU to ID now.

puntope (approved):
✅  LGTM thanks for this work @martynmjones 

Tested:

- Product SKU by default when `woocommerce_google_analytics_settings`is in Wp_Options
- Product ID when defaults when  `woocommerce_google_analytics_settings`is not in Wp_Options
- Product SKU shows in GA Debug
- Product ID shows in GA Debug
- Product #ID shows in GA Debug when no SKU and SKU option is selected

I left some questions and comments in the code. But Approved in advance.

woocommerce-google-analytics-integration.php
martynmjones: As part of the project to remove Universal Analytics we're also moving to using the product ID as the standard identifier (peeuvX-SJ-p2). One reason is to allow merchants using GLA to have the same product identification across both extensions by using the `woocommerce_ga_product_identifier` filter to prefix the ID with `gla_`.

Aside from that, the product ID is a constant and reliable way to identify a product whereas the SKU + fallback is more likely to cause issues for example if a merchant sets a SKU for an existing product or changes it then it would now be a "new" product in Google Analytics and not properly link up with historical data.

includes/class-wc-abstract-google-analytics-js.php
martynmjones: Nice catch! Update here 4371430

includes/class-wc-abstract-google-analytics-js.php
martynmjones: As the value won't change and how things are written elsewhere in the extension I've opted to leave as is for the sake of simplicity and readability.

martynmjones:
Many thanks for the review and input @puntope 🙌

PR #311, opened 2023-10-05
Update tracking for blocks

### LINKED ISSUE #268: [Follow-Up] - Compatibility with GA4
This issue is a follow-up for the related work in migrating the actions from WC Blocks to the extension.

However, GA4 did some changes that should be reflected in the code. 

GA4 drops the `event_category` and `event_label` structure in favor of individual parameters for each event. Ref: https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#view_item_list

_Originally posted by @martynmjones in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/264#discussion_r1177792610_

Another task is the one mentioned here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/264#discussion_r1177796264

And finally this one: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/264#discussion_r1177790632

⚠️  The standard plugin events (outside the actions) are also using the Universal format. So it will be needed to revisit that also.

martynmjones:

Closes #268

Tracking for blocks only supports Universal Analytics as it currently stands. This PR updates the tracking for GA4 and removes references to UA documentation.

### Screenshots:

Screenshot 2023-10-25 at 19 35 45

### `view_item_list`

1. Create a page using the `All Products` block and view page in incognito with GA Debug enabled (_Alternatively use Tag Assistant to monitor events_)
2. Confirm that on the initial page load a `view_item_list` event is triggered with the [correct parameters](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#view_item_list)
3. Click on the second page of the product listings and confirm that only one new `view_item_list` is triggered and that it matches the above data structure.

### `add_to_cart`

1. On the `All Products` page, click `Add to cart` for a product
2. Confirm that an `add_to_cart` event is triggered with the [correct parameters](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#add_to_cart)

### `remove_from_cart` and `change_cart_quantity`

- These events can be ignored with this PR. The experimental hooks in blocks don't work well for how GA4 expects the `remove_from_cart` event to work. #312 has been opened so that this can be addressed in a follow-up PR.

### `begin_checkout`

**Note:** The `experimental__woocommerce_blocks-checkout-render-checkout-form` hook which the `begin_checkout` event relies on doesn't currently work. Reported in https://github.com/woocommerce/woocommerce-blocks/issues/11472.

1. To test this event, go to the block-powered checkout page, apply a coupon, and run the following in console:
```js
wp.hooks.doAction('experimental__woocommerce_blocks-checkout-render-checkout-form', { storeCart: wp.data.select('wc/store/cart').getCartData() } );
```
2. Confirm that the `begin_checkout` event is triggered with the [correct parameters](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#begin_checkout) and that item quantities and coupon name are correct in the event data

### `add_shipping_info`

1. Add more than one shipping option to the test site
2. Select an option during checkout and click `Place Order`
2. Confirm that the `add_shipping_info` event is triggered with the [correct parameters](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#add_shipping_info)

### `select_content`

1. Go to a page with the `All Products` block
2. Click on a product
2. Confirm that the `select_content ` event is triggered with the [correct parameters](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#select_content)

### `search`

- Doesn't work at the moment. Also reported in https://github.com/woocommerce/woocommerce-blocks/issues/11472.

### Changelog entry

> Update - Tracking for GA4 when using Blocks

### DIFF
--- a/assets/js/src/tracking.js
+++ b/assets/js/src/tracking.js
@@ -89,8 +89,8 @@ export const trackBeginCheckout = ( { storeCart } ) => {
 			storeCart.totals.total_price,
 			storeCart.totals.currency_minor_unit
 		),
-		coupon: storeCart.coupons[ 0 ]?.code || '',
 		items: storeCart.items.map( getProductFieldObject ),
+		...( storeCart.coupons[ 0 ]?.code ? { coupon: storeCart.coupons[ 0 ]?.code } : {} ),
 	} );
 };
 
@@ -107,12 +107,12 @@ export const trackShippingTier = ( { storeCart } ) => {
 			storeCart.totals.total_price,
 			storeCart.totals.currency_minor_unit
 		),
-		coupon: storeCart.coupons[ 0 ]?.code || '',
 		shipping_tier:
 			storeCart.shippingRates[ 0 ]?.shipping_rates?.find(
 				( rate ) => rate.selected
 			)?.name || '',
 		items: storeCart.items.map( getProductFieldObject ),
+		...( storeCart.coupons[ 0 ]?.code ? { coupon: storeCart.coupons[ 0 ]?.code } : {} ),
 	} );
 };
 
--- a/assets/js/src/tracking.js
+++ b/assets/js/src/tracking.js
@@ -45,11 +45,6 @@ export const trackListProducts = ( {
  */
 export const trackAddToCart = ( { product, quantity = 1 } ) => {
 	trackEvent( 'add_to_cart', {
-		event_category: 'ecommerce',
-		event_label: __(
-			'Add to Cart',
-			'woocommerce-google-analytics-integration'
-		),
 		items: [ getProductFieldObject( product, quantity ) ],
 	} );
 };
--- a/assets/js/src/tracking.js
+++ b/assets/js/src/tracking.js
@@ -4,6 +4,7 @@ import {
 	getProductImpressionObject,
 	getProductId,
 	formatPrice,
+	getCartCoupon,
 } from './utils';
 
 /**
@@ -89,8 +90,8 @@ export const trackBeginCheckout = ( { storeCart } ) => {
 			storeCart.totals.total_price,
 			storeCart.totals.currency_minor_unit
 		),
+		...getCartCoupon( storeCart ),
 		items: storeCart.items.map( getProductFieldObject ),
-		...( storeCart.coupons[ 0 ]?.code ? { coupon: storeCart.coupons[ 0 ]?.code } : {} ),
 	} );
 };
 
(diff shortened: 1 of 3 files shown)


assets/js/src/tracking.js
```diff
@@ -23,17 +17,19 @@ export const trackListProducts = ( {
...
-			'woocommerce-google-analytics-integration'
-		),
-		items: products.map( ( product, index ) => ( {
-			...getProductImpressionObject( product, listName ),
-			list_position: index + 1,
-		} ) ),
-	} );
+	if ( products.length > 0 ) {
+		trackEvent( 'view_item_list', {
+			item_list_id: 'engagement',
```
tomalec, line 23: :scroll: :nail_care: 
If I understand the docs correctly, this id is to identify the list from which we're reporting.
With blocks, we can have multiple lists on different pages. But currently, we send `engagement` regardless of which list we track. Maybe we could somehow forward the block instance id, or at least page ID?

assets/js/src/tracking.js
```diff
@@ -91,70 +77,55 @@ export const trackChangeCartItemQuantity = ( { product, quantity = 1 } ) => {
...
-
-		currentStep = step;
-	};
+export const trackBeginCheckout = ( { storeCart } ) => {
+	trackEvent( 'begin_checkout', {
+		currency: storeCart.totals.currency_code,
+		value: formatPrice(
+			storeCart.totals.total_price,
+			storeCart.totals.currency_minor_unit
+		),
```
tomalec, line 92: Documentation [states](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#parameters_5) it should be a `number`, we send a `string`.

assets/js/src/tracking.js
```diff
@@ -91,70 +77,55 @@ export const trackChangeCartItemQuantity = ( { product, quantity = 1 } ) => {
...
+		coupon: storeCart.coupons[ 0 ]?.code || '',
```
tomalec: :grey_question:  
I wonder if the empty `""` in case of lack of coupon is the most accurate representation. Shouldn't we just don't send this property in case of lack of coupon? I'm not sure how the value being set (event to empty string) would be interpreted.

tomalec:
The Google documentation states that the price parameter should be a number, but we're sending string
image

tomalec (approved):
Reviewed the code tested locally.
My only concern is the price data type.
Other than that, LGTM.

assets/js/src/tracking.js
martynmjones: That would be a nice addition! I've opened issue #324 so as to not get off track in this PR.

assets/js/src/tracking.js
martynmjones: Nice catch! Updated in 4a06f52

assets/js/src/tracking.js
martynmjones: That's a good point. I'm not sure what the behavior would be and from a quick search I couldn't find any thing in the GA documentation about it 🤔 I've updated how it works in 813b7ee so we're not including an empty string.

PR #313, opened 2023-10-06
Move tracking for classic pages to JavaScript

martynmjones:

Closes # .

_Replace this with a good description of your changes & reasoning._

### Checks:

1. 
2. 
3. 

### Changelog entry

>

### DIFF
--- a/assets/js/src/actions.js
+++ b/assets/js/src/actions.js
@@ -1,5 +1,5 @@
 import { removeAction } from '@wordpress/hooks';
-import { NAMESPACE, ACTION_PREFIX } from './constants';
+import { NAMESPACE, ACTION_PREFIX, EVENT_PREFIX } from './constants';
 import {
 	trackBeginCheckout,
 	trackShippingTier,
@@ -12,7 +12,7 @@ import {
 	trackViewItem,
 	trackException,
 } from './tracking';
-import { addUniqueAction } from './utils';
+import { addUniqueAction, eventListener } from './utils';
 
 /**
  * Track begin_checkout
@@ -71,6 +71,8 @@ addUniqueAction(
 	NAMESPACE,
 	trackAddToCart
 );
+
+eventListener( `${ EVENT_PREFIX }-add_to_cart`, trackAddToCart );
 
 /**
  * Change cart item quantities
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -102,8 +102,6 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 	 * @param array  $cart_item_data extra cart item data we want to pass into the item.
 	 *
 	 * @return void
-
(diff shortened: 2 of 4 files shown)


assets/js/src/actions.js
```diff
@@ -72,6 +72,8 @@ addUniqueAction(
 	trackAddToCart
 );
 
+eventListener( `${ EVENT_PREFIX }-add_to_cart`, trackAddToCart );
+
```
tomalec, line 76: Is there a reason to use jQuery here? I think we can reduce the size of dependency and bloat by using native JS
```js
document.body.addEventListener(
    `${ EVENT_PREFIX }-add_to_cart`,
    event =>  trackAddToCart( {product: event.detail } )
);
// ...
document.body.dispatchEvent(
    new CustomEvent( 
        `google_analytics-add_to_cart`, 
        {detail: {foo: 'bar'}}
    )
);
     

```

includes/class-wc-google-gtag-js.php
```diff
@@ -507,4 +572,22 @@ public function event_tracking_code( $parameters, $selector ) {
...
+	 * @param array  $event_data Data to include with the event
+	 *
+	 * @return void
+	 */
+	public function trigger_event( $event_name, $event_data ) {
+		wc_enqueue_js(
+			sprintf(
+				"$( document.body ).trigger( 'google_analytics-%s', '%s' )",
+				esc_js( $event_name ),
+				wp_json_encode( $event_data )
```
tomalec, line 588: Also, can we call the `trackAddToCart` function synchronously? Instead of bubbling events?

I can imagine we can do so with
1. publishing the extension's namespace in the global scope for some shared stuff
```php
"window['woocommerce-google-analytics-integration'].trackEvent('%s', '%s' )",
esc_js( $event_name ),
wp_json_encode( $event_data )
```
2. attaching the extension's shared namespace to its own script tag. Script tag, whose handle we control
```php
"document.getElementById('%s').wcgai.trackEvent('%s', '%s' )",
$this->script_handle . '-actions-js',
esc_js( $event_name ),
wp_json_encode( $event_data )

// in actions.js
document.currentScript.wcgai = { //...
```

Each solution - yours included - relies on squatting a unique name in the public, global space. So I'd say each is equally bad in scope pollution metric. So we should evaluate other aspects :;)
1. We introduce a global variable - which makes things least magical and easy to inspect
2. We rely on the handle id that needs to be there for other reasons. So, theoretically, we do not pollute more than before. However, it's slightly tricky the way we access the scope.
3. Event-based solution. Relies on a unique event name. Not so straightforward and inspectable. For example. I imagine we can face a racy condition when we fire an event before registering the listener, then it would be hard to notice that something went wrong. Logs will be clean for HEs to check.

martynmjones:
This branch and PR was primarily for exploration and after discussion with @tomalec we will be going in a slightly different direction with how the tracking will work so I'm closing this PR for now.

PR #316, opened 2023-10-10
Declare Compatibility for cart_checkout_blocks

puntope:

This PR marks the plugin as compatible with the cart_checkout_blocks feature.

n/a

*At the moment seems like no feature depends on the compatibility to be set as TRUE.
Possible to verify tho checking the instructions here pdFofs-1qY-p2

### Changelog entry

> Tweak - Declare cart_checkout_blocks feature compatibility


jorgemd24 (approved):
Thanks, @puntope, LGTM. I did some tests by simply toggling between the values `true` and `false` in the filter to ensure it doesn't appear as incompatible. Similar to what you mentioned in this PR: https://github.com/woocommerce/automatewoo-referrals/pull/244.

PR #317, opened 2023-10-10
Slightly simplify JS construction in WC_Google_Gtag_JS

tomalec:

These are a few cosmetic changes I made trying to understand what code is being constructed by PHP. That should be fully backward compatible, but makes the code slightly more readable at least to me.
- [Simplify JS construction in WC_Google_Gtag_JS::event_tracking_code](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/c77cbf124359401abb1abca630e95fc3b88111a1)
- [DRY JS construction in WC_Google_Gtag_JS::listing_click](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/1c5010de01cf421326a4403d62fa870da0c8dcb7)
- [Simplify gtag installatio script.](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/3dcc5246289c2b2cfa51dff78d815b0c40ab853d) 
   Use `gtag` locally.
   

### Checks:

1. Review the code generated from `listing_click` and `event_tracking_code` - it should be the same (ignoring whitespace)
2. Review the code generated from `load_analytics`. It should be effectively equivalent

### DIFF
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -205,7 +205,7 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 				'name'     => $product->get_title(),
 				'category' => self::product_get_category_line( $product ),
 				'quantity' => 1,
-			]
+			],
 		];
 
 		wc_enqueue_js(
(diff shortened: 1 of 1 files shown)


includes/class-wc-google-gtag-js.php
```diff
@@ -286,23 +278,26 @@ public static function load_analytics( $order = false ) {
...
-		$gtag_snippet       = '
+
+		$custom_gtag = self::tracker_var();
+		$gtag_snippet        = "(()=>{
 		window.dataLayer = window.dataLayer || [];
-		function ' . self::tracker_var() . '(){dataLayer.push(arguments);}
-		' . self::tracker_var() . "('js', new Date());
+		// Publish gtag under custom name to the global scope.
+		// Preserve the custom `arguments.callee.name`, for backward compatibility (not sure if even needed).
+		const gtag = window." . $custom_gtag . " = function " . $custom_gtag . "(){dataLayer.push(arguments);}
```
tomalec, line 297: 1. I'm not sure if we or Google are bothered about `arguments.callee.name` that's pushed to `window.dataLayer`. If we don't mind pushing always `gtag` regardless of the custom name in `tracker_var`, we could simplify it to just `window.
```suggestion
		window." . $custom_gtag . " = function gtag(){dataLayer.push(arguments);}
```

tomalec, line 297: Or if the `tracker_var` was to just avoid conflicts with other plugins using the "gtag", and not to alow customization of something veryspecific. We can hide it in our own namespace/scope. Like
```suggestion
		window.wcgai.gtag = function gtag(){dataLayer.push(arguments);}
```
And use it as `wcgai.gtag( 'even', //...` everywhere else

includes/class-wc-google-gtag-js.php
```diff
@@ -286,23 +278,26 @@ public static function load_analytics( $order = false ) {
...
 		$gtag_cross_domains = ! empty( self::get( 'ga_linker_cross_domains' ) ) ? array_map( 'esc_js', explode( ',', self::get( 'ga_linker_cross_domains' ) ) ) : array();
-		$gtag_snippet       = '
+
+		$gtag_snippet = "(()=>{
 		window.dataLayer = window.dataLayer || [];
-		function ' . self::tracker_var() . '(){dataLayer.push(arguments);}
-		' . self::tracker_var() . "('js', new Date());
+		// Publish gtag under custom name to the global scope.
+		// Preserve the custom `arguments.callee.name`, for backward compatibility (not sure if even needed).
+		const gtag = window.$gtag_name = function $gtag_name(){dataLayer.push(arguments);}
```
tomalec, line 299: 1. I'm not sure if we or Google are bothered about `arguments.callee.name` that's pushed to `window.dataLayer`. If we don't mind pushing always `gtag` regardless of the custom name in `tracker_var`, we could simplify it to just `window.
```suggestion
		window." . $custom_gtag . " = function gtag(){dataLayer.push(arguments);}
```

tomalec, line 299: 2. Or if the `tracker_var` was just to avoid conflicts with other plugins using the "gtag", and not to allow customization of something very specific. We can hide it in our own namespace/scope. Like
```suggestion
		window.wcgai.gtag = function gtag(){dataLayer.push(arguments);}
```
And use it as `wcgai.gtag( 'event', //...` everywhere else

tomalec:
Actually, I think we can also merge this change to the develop, as it is not related to UA removal, and should work as well there.
An maybe it will reduce merge conflicts if UA removal branch is going to take some time.

martynmjones (commented):
Hey @tomalec, thanks for your work to simplify the tracker.

I'm wondering if there's any benefit to us continuing to setup the tracker the way we are though? As we discussed on the call a couple of weeks back, we're going to remove as much inline JS as possible and instead trigger [function calls](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/04e883a8dcf9b72269ff8244d3a70bb5622b6eba/assets/js/src/tracking.js) with relevant data.

In `class-wc-google-gtag-js.php` we could setup an array with all relevant settings and then trigger a call to a JS function that initializes the tracker function which would make `load_analytics` a lot simpler and far more readable.

What do you think?

tomalec:
I totally agree :)
I just had these bits as I was trying to understand the code, so I thought It might be a useful first step in the process of refactoring, or we could push it in the meantime before UA removal if that's going to take more time.
As well, I'm happy to discard this PR completely in favor of deeper refactoring.

martynmjones:
> I just had these bits as I was trying to understand the code, so I thought It might be a useful first step in the process of refactoring, or we could push it in the meantime before UA removal if that's going to take more time.
As well, I'm happy to discard this PR completely in favor of deeper refactoring.

Definitely useful and thanks for opening the PR!

As we are going to be making some pretty significant changes I'd be in favor of jumping straight into a deeper refactoring of how it's working at the moment.

While the project will take a while longer, I'm not sure I see the benefit in pushing to `trunk` and releasing as the extension does work at the moment it's just not particularly clean. That being said, if you disagree then let me know and I'll happily run further tests and review properly so that it can be merged.

tomalec:
closing in favor of https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328

PR #320, opened 2023-10-27
Remove Universal Analytics code

martynmjones:

Removes all existing Universal Analytics code from the extension. 

1. Review the changes and deleted code
2. Run some very basic smoke tests to confirm no errors are triggered by the changes

### Additional details:

In addition to removing the UA code, some tracking was moved from `class-wc-google-analytics.php` c566ea7 into `class-wc-google-gtag-js.php` 2ee2fad. The tracking itself doesn't need testing here as it will be completely reworked in a follow up PR.

### Changelog entry

> Remove - Universal Analytics code


includes/class-wc-google-analytics.php
```diff
@@ -383,13 +300,10 @@ public function track_options( $data ) {
...
 			'support_enhanced_link_attribution'   => $this->ga_support_enhanced_link_attribution,
-			'use_universal_analytics'             => $this->ga_use_universal_analytics,
 			'anonymize_enabled'                   => $this->ga_anonymize_enabled,
 			'ga_404_tracking_enabled'             => $this->ga_404_tracking_enabled,
 			'ecommerce_tracking_enabled'          => $this->ga_ecommerce_tracking_enabled,
 			'event_tracking_enabled'              => $this->ga_event_tracking_enabled,
-			'gtag_enabled'                        => $this->ga_gtag_enabled,
-			'set_domain_name'                     => empty( $this->ga_set_domain_name ) ? 'no' : 'yes',
 			'plugin_version'                      => WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION,
 			'enhanced_ecommerce_tracking_enabled' => $this->ga_enhanced_ecommerce_tracking_enabled,
```
tomalec: I think we should remove this as well.
```suggestion
```
or set it to static `null`

tomalec: Maybe also with https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/320/files#diff-75fada6b677c39f17a7de173a315db4d8911e7469f5a70dce4822a6aaa849514R551-R566
I guess `event_tracking_code` is about to be rewritten anyway, but we can get rid of dead-code earlier to make remove any confusion from following PRs

includes/class-wc-google-gtag-js.php
```diff
@@ -263,7 +276,7 @@ public static function add_to_cart( WC_Product $product ) {
 		}
 
 		$event_code .= self::get_event_code(
-			'add_to_cart',
+			'add_to_cart_x',
```
tomalec: :question: 
Is that `_x` here on purpose? or just a leftover from debugging?
```suggestion
			'add_to_cart',
```

includes/class-wc-google-analytics.php
```diff
@@ -536,193 +422,6 @@ private function disable_tracking( $type ) {
...
-		}
-
-		$url = str_replace( 'href=', 'data-product_id="' . esc_attr( $product->get_id() ) . '" data-product_sku="' . esc_attr( $product->get_sku() ) . '" href=', $url );
-		return $url;
-	}
-
-	/**
-	 * Google Analytics event tracking for loop add to cart
-	 */
-	public function loop_add_to_cart() {
```
tomalec, line 629: :grey_question:  We didn't move it to the `gtag-js.php`.  When and where it was actually called, what was the reason to call it on `wp_footer`?

includes/class-wc-google-analytics.php
```diff
@@ -536,193 +422,6 @@ private function disable_tracking( $type ) {
...
-
-	/**
-	 * Measure a product click and impression from a Product list
-	 *
-	 * @param string     $link The Add To Cart Link
-	 * @param WC_Product $product The Product
-	 */
-	public function track_product( $link, $product ) {
-		if ( $this->enhanced_ecommerce_enabled( $this->ga_enhanced_product_click_enabled ) ) {
-			$this->get_tracking_instance()->listing_impression( $product );
```
tomalec, line 694: :grey_question: 
Was that a bug that we were tracking `impression` when `ga_enhanced_product_click_enabled` is set regardless of `ga_enhanced_product_impression_enabled`, or do I miss something more?

tomalec:
As we removed the "Enhances e-Commerce" section, I think we should move its options to the main section (remove `				'class'         => 'enhanced-setting'`). 
Currently, it seems like they are a part of "Cross Domain Tracking"
image

includes/class-wc-google-gtag-js.php
martynmjones: Debugging 🤦 many thanks for catching! fe0ced4

includes/class-wc-google-analytics.php
martynmjones: Removed the redundant tracks data and re-ordered settings 2b48a17 along with removing the deprecated function a7e9ac4

includes/class-wc-google-analytics.php
martynmjones: The `loop_add_to_cart` function was only used for Universal Analytics and if gtag was enabled it would [immediately return](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/39757b6914d7bbeae1d1394e09f32fffd39eaf41/includes/class-wc-google-analytics.php#L619). The add to cart tracking will likely change quite a bit but for now the aim is just to remove all UA code.

includes/class-wc-google-analytics.php
martynmjones: Yeah that was a bug which meant if tracking product clicks was enabled then we would also track impressions. Again this will probably change quite a bit but for now it's just a simple bugfix.

martynmjones:
Many thanks for the review @tomalec! I've addressed the questions and points you raised so this is ready for a second look 👀.

tomalec (approved):
Thanks for the changes and clarifications, LGTM

PR #321, opened 2023-10-27
Add upgrade notice if property ID starts with "UA"

martynmjones:

Adds an admin notice if the stored property ID begins with "UA" prompting merchants to use Google's setup assistant and then update their WooCommerce settings.

### Screenshots:

Screenshot 2023-10-27 at 17 04 01

1. Checkout `add/ua-update-notice` on a site with the `Google Analytics Tracking ID` set to `UA-.....`
2. Visit other admin areas to confirm the notice is displayed
3. Confirm notice is clear and links work
4. Update property ID to start with `G-`
5. Confirm notice is no longer displayed

### Changelog entry

> Add - Update notice for merchants using a Universal Analytics Property ID

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -90,6 +90,8 @@ class WC_Google_Analytics extends WC_Integration {
 		$this->init_settings();
 		$constructor = $this->init_options();
 
+		add_action( 'admin_notices', array( $this, 'universal_analytics_upgrade_notice' ) );
+
 		// Contains snippets/JS tracking code
 		include_once 'class-wc-abstract-google-analytics-js.php';
 		include_once 'class-wc-google-gtag-js.php';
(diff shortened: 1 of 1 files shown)


includes/class-wc-google-analytics.php
```diff
@@ -111,6 +113,26 @@ public function __construct() {
...
+	 * @return void
+	 */
+	public function universal_analytics_upgrade_notice() {
+		if ( 'ua' === substr( strtolower( $this->get_option( 'ga_id' ) ), 0, 2 ) ) {
+			echo sprintf(
+				'<div class="%1$s"><p>%2$s</p></div>',
+				'notice notice-error',
+				sprintf(
+					/* translators: 1) URL for Google documentation on upgrading from UA to GA4 2) URL to WooCommerce Google Analytics settings page */
+					__( 'Your website is configured to use Universal Analytics which Google retired in July of 2023. Update your account using the <a href="%1$s" target="_blank">setup assistant</a> and then update your <a href="%2$s">WooCommerce settings</a>.', 'woocommerce-google-analytics-integration' ), // phpcs:ignore WordPress.Security.EscapeOutput
```
ianlin: I'm thinking if using `esc_html__` would be better even though we are not using any variables in the string. I slightly lean towards to escape it like the code below to show privacy policy so it would be more consistent and also it could keep the string safe if we were to use the variables inside it in the future.

https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/fd97569b48b17a126541bb62096b74a3f61ccafa/includes/class-wc-google-analytics.php#L432-L437

includes/class-wc-google-analytics.php
```diff
@@ -111,6 +113,26 @@ public function __construct() {
...
+					'https://support.google.com/analytics/answer/9744165?sjid=9632005471070882766-EU#zippy=%2Cin-this-article',
```
ianlin: ❔ Does the parameter `sjid=9632005471070882766-EU` has a special meaning? If not I think we could just use the link https://support.google.com/analytics/answer/9744165.

ianlin (commented):
Thanks for the change, it worked as expected. I can see the notice being shown when the property ID set to `UA-xxx`. I just left some questions that I wasn't very sure for discussion.

includes/class-wc-google-analytics.php
martynmjones: Thanks for picking up on that!

includes/class-wc-google-analytics.php
martynmjones: Updated in 1f0b6fc

martynmjones:
Hey @ianlin, many thanks for the review! I've made the changes you suggested so this is ready for another look 👀

ianlin (approved):
Thanks for the extra changes, LGTM 👍

PR #328, opened 2023-11-10
Update tracking for classic pages

martynmjones:

This is a major restructuring centered around a new [Tracker utility](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/66142e273047f086bf00a780d72bd0b1cb1e5781/assets/js/src/tracker/index.js) which manages all Google Analytics events triggered by either [WooCommerce Blocks](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/66142e273047f086bf00a780d72bd0b1cb1e5781/assets/js/src/integrations/blocks.js) or [classic pages](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/66142e273047f086bf00a780d72bd0b1cb1e5781/assets/js/src/integrations/classic.js). All tracked events must have an accompanying [data formatter](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/66142e273047f086bf00a780d72bd0b1cb1e5781/assets/js/src/tracker/data-formatting.js) in order for the Tracker to send the event.

Several files and tests have been removed which are no longer of use 

### Screenshots:

Screenshot 2024-01-25 at 17 36 37

### Test setup

Debug a test website via [Tag Assistant](https://tagassistant.google.com/) that has both Block powered and classic pages for the Cart, Checkout, and Product Listings (Using the `All Products` Block). Ensure all `Event Tracking` options are enabled in `WooCommerce > Settings > Integrations > Google Analytics`.

_Work through all test instructions using both types of pages while not logged into the site as an admin._

1. Visit a product archive page and confirm that **one** `view_item_list` event is sent containing all the products on the page
2. Click on a product
3. Confirm that a `select_content` event is sent for that product
4. After the single product loads, confirm that a `view_item` event is sent for the product and a `view_item_list` event is sent if any related products or other lists are displayed
5. Add the product to cart and confirm the `add_to_cart` event is sent
6. Visit the cart, remove the product from the cart and confirm the `remove_from_cart` event it sent
7. Go back to the product archive page and add products to cart from there confirming the event is sent for each
8. Visit the checkout and confirm a `begin_checkout` event is sent
9. Place the order and one the thank you page confirm a `purchase` event is sent
10. Run some basic smoke tests

#### Additional

1. Disable some options in `WooCommerce > Settings > Integrations > Google Analytics`
2. Confirm those events are no longer sent
3. Adjust the `Product Identification` settings
4. Confirm that the data sent with each event includes the correct product identifier structure

⚠️ This branch also includes the changes needed to rename the extension to `Google Analytics for WooCommerce` (_See #353_).

### Changelog entry

> Update - Add shared tracking functionality for WooCommerce Blocks and classic pages


assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,104 @@
...
+				true
+			);
+			window[ config.tracker_var ]( 'config', config.gtag_id, {
+				allow_google_signals: config.allow_google_signals,
+				link_attribution: config.link_attribution,
+				anonymize_ip: config.anonymize_ip,
+				logged_in: config.logged_in,
+				linker: config.linker,
+				custom_map: config.custom_map,
+			} );
```
tomalec: :nail_care: WDYT of:?

```suggestion
			function gtag() {
				window.dataLayer.push( arguments );
			};
			window[ config.tracker_var ] = gtag;
			gtag( 'js', new Date() );
			gtag(
				'set',
				`developer_id.${ config.developer_id }`,
				true
			);
			gtag( 'config', config.gtag_id, {
				allow_google_signals: config.allow_google_signals,
				link_attribution: config.link_attribution,
				anonymize_ip: config.anonymize_ip,
				logged_in: config.logged_in,
				linker: config.linker,
				custom_map: config.custom_map,
			} );
```

assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,104 @@
...
+ * @return {Object} An object containing methods: init, setupEvents, event{ attach, get, track }.
+ */
+export const tracker = ( () => {
+	const events = [];
+
+	return {
+		/**
+		 * Initializes the tracker and dataLayer if not already done.
+		 */
+		init: () => {
```
tomalec: :nail_care: 
Instead of object literal and `init`, I'd use class syntax to create a singleton, something like
```js
class Tracker{
	/**
	 * Initializes the tracker and dataLayer if not already done.
	 */
	constructor() {
	  // We can protect against `new tracker.constructor()` if we want to
	  if(instance){
	  	throw new Error("New instance cannot be created!!");
	  	// or more gently, but a bit tricky
	        return instance;
	    }
	    instance = this;
//...

export const tracker = Object.freeze(new Tracker());
```

Then we can add regular methods and properties. Allowing us to reason and inspect the object more structurally.

assets/js/src/tracker/index.js
tomalec: Or if we really want to keep the `window.dataLayer[0].callee.name === ''` (correct me if I'm wrong, but I don't think we have such a need)
```js
const gtag = window[ config.tracker_var ] = function() {
```

assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,104 @@
...
+		},
+		/**
+		 * An object to control events which exist in the global `events` object.
+		 *
+		 * @param {string} event - The name of the event.
+		 *
+		 * @return {Object} An object containing methods to attach, get, and track the event.
+		 */
+		event: ( event ) => {
+			return {
```
tomalec: :question: 
What's the purpose of creating this object every time we call `tracker.event`?
I'm not sure I follow the logic and reasoning here, plus I'm overwhelmed a bit with the number of things called "event" ;)
(global `events` array, `tracker.event`method , its `event` param, then `eventObject`s used bellow, etc.) I think we can reduce it a bit.

assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,104 @@
...
+				/**
+				 * Attach data to an event and track it.
+				 *
+				 * @param {*} data - Data to be used for the event.
+				 */
+				attach: ( data ) => {
+					const eventObject = tracker.event( event ).get();
```
tomalec: :question: 
as this `attach` methods is already a property of the object created by `tracker.event(event)`
I
```suggestion
					const eventObject = this.get();
```

assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,104 @@
...
+					const eventData = eventObject.callback( data );
```
tomalec: :construction: 
Theoretically `eventObject` can be `false`, then it will throw.

assets/js/src/tracker/index.js
tomalec: For example, does something like bellow is missing something we need:
```js
tracker.attachEvent( eventName ) = ( data ) => {
```

tomalec (commented):
I like the idea I get from the brief description of the approach in the P2.
However, I'm not sure I fully understand specifics and implementation.

I tried to simplify the implementation logic a bit, maintaining more-or-less the same behavior.
By proposal is at https://github.com/woocommerce/woocommerce-google-analytics-integration/compare/update/tracking-for-classic-pages...update/tracking-for-classic-pages-tracker-class?expand=1

Please, let me know if that makes sense to you and still goes in the direction you envision.

I think there is still room for simplification in behavior and logic in both branches :)

I'd appreciate a F2F call, to discuss the details.

assets/js/src/tracker/data-formatting.js
```diff
@@ -154,9 +156,9 @@ export const trackViewItem = ( {
 	listName = __( 'Product List', 'woocommerce-google-analytics-integration' ),
 } ) => {
 	if ( product ) {
-		trackEvent( 'view_item', {
+		return {
 			items: [ getProductImpressionObject( product, listName ) ],
-		} );
+		};
 	}
 };
```
tomalec, line 182: Some of the functions here still call `trackEvent`. Do I understand correctly, that it's because it's WIP, and eventually all functions here will only format/translate the data? Then the actual "tracking" will be covered by `tracker…track()`?

If so, maybe we could rename them from `track*` to `format*`?

assets/js/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Register all Google Analytics 4 events that can be tracked
+ */
+tracker.setupEvents( [
+	{
+		/**
+		 * @see https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#begin_checkout
+		 */
+		name: 'begin_checkout',
+		callback: trackBeginCheckout,
```
tomalec: :nail_care: 
I got a feeling we're adding more work for ourselves by translating it over and over.
1. `./tracker/data-formatting.js` defines a set of functions `track*`
2. here we map event name like "begin_checkout" to "trackBeginCheckout"
3. `./tracker/index.js` defines a class and implements methods to retrieve `track*` for given name.
4. `integrations/blocks.js` maps block events to our event names
5. `includes/class-wc-google-gtag-js.php` maps WC hook names to our event names

We have 2 disjointed sets of names given by the platform
- WC hooks
- black events
So we introduce our set of `event_names` to handle them. I don't see a reason to introduce fourth set of `trackFunctionNames`.

What I'd propose is to do something like
```js
// tracker/index.js
import * as eventDataFormats from './data-formatting.js';
// …
attach( eventName ) {
  return (data) => this.track( 'event', eventName, eventDataFormats[ eventName ]( data ) );
}
```

and `data-formatting.js` will be the only source of our event names and respective formatters

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,106 @@
...
+		?.forEach( ( button ) => {
+			const productId = button.dataset.product_id;
+
+			button.parentNode.addEventListener( 'click', ( listing ) => {
+				const targetLink = listing.target.closest(
+					'.woocommerce-loop-product__link'
+				);
+				const isAddToCartButton =
+					button.classList.contains( 'add_to_cart_button' ) &&
+					! button.classList.contains( 'product_type_variable' );
```
tomalec: :grey_question: Why do we check if that's an add to cart button after adding the listener? Can we check that before? and add listeners only to those? Or are those classes added/changed async?

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,106 @@
...
+						parseInt( item.dataset.product_id ),
+						true
+					),
+				} );
+			} );
+	};
+
+	removeFromCartListener();
+
+	document.body.onupdated_wc_div = () => removeFromCartListener();
```
tomalec: :question: :construction: This seems to possibly assign the listener twice. That could result in tracking the same action twice.

assets/js/src/integrations/blocks.js
```diff
@@ -0,0 +1,51 @@
...
+);
+
+addUniqueAction(
+	`${ ACTION_PREFIX }-product-view-link`,
+	NAMESPACE,
+	tracker.event( 'select_content' ).handler
+);
+
+/**
+ * Temporarily remove all actions for demo purposes.
```
tomalec: Is this comment still relevant?

assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,81 @@
...
+			throw new Error( `Event ${ name } is not supported.` );
+		}
+
+		return {
+			handler: ( data ) => {
+				if ( config.events.includes( name ) ) {
+					window[ config.tracker_var ](
+						'event',
+						name,
+						formatters[ name ]( data )
```
tomalec: We're checking the presence of `formatters[ name ]` function at the time of `event()` call, but we use it much later, when technically, the function may no longer exist (probably unexpected, but technically possible).
In my opinion, we could make it a bit safer: we could cache the function once it is checked. So, the handler will call the function available at the `event` execution time:
```suggestion
		/* eslint import/namespace: [ 'error', { allowComputed: true } ] */
		const formatter = formatters[ name ];
		if ( typeof formatter !== 'function' ) {
			throw new Error( `Event ${ name } is not supported.` );
		}

		return {
			handler: ( data ) => {
				if ( config.events.includes( name ) ) {
					window[ config.tracker_var ](
						'event',
						name,
						formatter( data )
```

assets/js/src/tracker/index.js
```diff
@@ -0,0 +1,81 @@
...
+	 * @return {{handler: function(*): void}} An object with a `handler` method for processing and tracking the event.
+	 * @throws {Error} If the event name is not supported.
+	 */
+	event( name ) {
+		/* eslint import/namespace: [ 'error', { allowComputed: true } ] */
+		if ( ! formatters[ name ] ) {
+			throw new Error( `Event ${ name } is not supported.` );
+		}
+
+		return {
```
tomalec: We don't need this object, do we?

Currently, every use of it is like: `tracker.event( … ).handler…`. The event object itself is not used. So, I think we can remove it, and simplify the APIs and signatures.

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,106 @@
...
+	const eventData = {
+		storeCart: cart,
+		products,
+		product,
+		order,
+	};
+
+	Object.values( events ?? {} ).forEach( ( eventName ) => {
+		if ( eventName === 'add_to_cart' ) {
+			tracker.event( eventName ).handler( { product: addedToCart } );
```
tomalec: :grey_question: Why do we use a different format for blocks and classic checkout for `add_to_cart`?

includes/class-wc-google-gtag-js.php
```diff
@@ -37,498 +36,185 @@ public static function get_instance( $options = array() ) {
...
-			function( &$value, $key ) {
-				$value = esc_js( $value );
+	public function map_actions(): void {
+		array_walk(
+			$this->mappings,
+			function( $hook, $gtag_event ) {
+				add_action(
+					$hook,
+					function() use ( $gtag_event ) {
+						$this->set_script_data( 'events', $gtag_event, $gtag_event );
```
tomalec: :grey_question: 
Is it correct?
<code>this->set_script_data( 'events', <strong>$gtag_event, $gtag_event</strong> );</code>

If I understood correctly, this would result in doing
```php
$this->script_data[ 'events' ][ 'begin_checkout' ] = 'begin_checkout'
```
on `woocommerce_before_checkout_form` hook.

So, we're sending event name as the event data.

includes/class-wc-google-gtag-js.php
```diff
@@ -37,498 +36,185 @@ public static function get_instance( $options = array() ) {
...
-	}
-
-	/**
-	 * Output Javascript to track add_to_cart event on single product page
+	 * @return void
 	 */
-	public static function add_to_cart() {
-		if ( 'yes' !== self::get( 'ga_event_tracking_enabled' ) || ! is_single() ) {
-			return;
+	public function set_script_data( string $type, $data, ?string $key = null, bool $single = false ): void {
```
tomalec: :nail_care: 
I find this API a bit confusing, as if I understood correctly it's like:
```
set( where1, what, [where2], [where_modifier])
```
WDYT of something like:

```
set( what, where1, [where2], [where_modifier])
```
so:
```suggestion
	public function set_script_data( $data, string $type, ?string $key = null, bool $single = false ): void {
```

tomalec: Also, the majority of useage does: `set_script_data( *, *, null, true)`, and it would behave the same like `set_script_data( *, *, "this-argument-is-pointless-now", true)`. 

So maybe we could make the "overwrite the root property" the default. Change the API, to something like:
`set_script_data( *, *, ?string $key_or_push = null ): void {`

```php
	/**
	 * Add an event to the script data
	 *
	 * @param string           $type The type of event this data is related to.
	 * @param string|array     $data The event data to add.
	 * @param string|bool|null $key  Set to add the data as a new array item, `true` to push with a numeric key.
	 *
	 * @return void
	 */
	public function set_script_data( string $type, $data, ?string $key = null ): void {
	// or public function set_script_data( $data, string $type, ?string $key = null ): void {
		if ( is_null( $key ) ) {
			$this->script_data[ $type ] = $data;
		} else {
			if ( ! isset( $this->script_data[ $type ] ) ) {
				$this->script_data[ $type ] = array();
			}
			if( $key === true ) {
				$this->script_data[ $type ][] = $data;
			} else {
				$this->script_data[ $type ][ $key ] = $data;
			}
		}
	}
```

Then the calls would look like:
- overwrite the root: 
   ```php
   $this->set_script_data( *, * );
   $this->set_script_data( 'cart', $this->get_formatted_cart() );
   ```
- specify key: 
   ```php
   $this->set_script_data( *, *, 'key' );
   $this->set_script_data( 'events', $gtag_event, $gtag_event );
   ```
- push to array: 
   ```php
   $this->set_script_data( *, *, true );
   $this->set_script_data( 'products', $this->get_formatted_product( $product ), true );
   ```

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -67,117 +141,190 @@ function gaOptout() {
...
 	}
 
 	/**
-	 * Echo JavaScript to track an enhanced ecommerce remove from cart action
+	 * Returns the tracker variable this integration should use
+	 *
+	 * @return string
 	 */
-	abstract public function remove_from_cart();
+	abstract public static function tracker_var(): string;
```
tomalec: :nail_care: 
I wonder if "tracker variable" is the best name here. Wouldn't it be confused with the "tracker" meaning the instance of Tracker JS class? I understand the current name comes from the old code, where it was to literally describe that PHP was defining JS `var`.


This value would be used as the "global tracking function name". So maybe something like:
- `tracker_function_name`
- `tracker_function`
- `tracker_name`
?

assets/js/src/config.js
```diff
@@ -0,0 +1,8 @@
+/* global wcgaiData */
+export const config = wcgaiData.config ?? {};
+export const events = wcgaiData.events ?? {};
+export const cart = wcgaiData.cart ?? {};
+export const products = wcgaiData.products ?? {};
+export const product = wcgaiData.product ?? {};
+export const addedToCart = wcgaiData.added_to_cart ?? {};
+export const order = wcgaiData.order ?? {};
```
tomalec: :nail_care: 
Could be rewritten to a tad shorter version.
It also minimizes a bit better 5 864b vs 6023b
```suggestion
/* global wcgaiData */
export const {
	config = {},
	events = {},
	cart = {},
	products = {},
	product = {},
	addedToCart: added_to_cart = {},
	order = {},
} = wcgaiData;
```

tomalec: :grey_question: :nail_care: 
 I also wonder, wouldn't it be safer and cleaner to leave `undefined` where it is?
If PHP didn't expose something, let further JS code be aware of it. Right now this file adds a masking layer. 
Further JS code can do `?.` or `Object.values( events ?? {} ).forEach(` as we do.

But if we avoid setting `= {}` here, we could have an insight in JS that PHP didn;t deliver something, and react accordingly.

tomalec (changes requested):
I reviewed the code and left a few comments. The major blocking one is the double event listener. https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328#discussion_r1473645572

I tested the functionality only briefly, I'll continue deeper tests later, but I'd also like to propose another change in a separate branch: https://github.com/woocommerce/woocommerce-google-analytics-integration/compare/update/tracking-for-classic-pages...update/tracking-for-classic-pages-simplified-tracker?expand=1

assets/js/src/tracker/index.js
tomalec: WDYT of something like: https://github.com/woocommerce/woocommerce-google-analytics-integration/compare/update/tracking-for-classic-pages...update/tracking-for-classic-pages-simplified-tracker?expand=1

assets/js/src/integrations/classic.js
martynmjones: Here we're adding an event listener to `.product`, getting the product ID from the data attribute and then sending a `select_content` event when a link is clicked on.

We need to add this check to make sure the event does fire if someone clicks on the main button when it's for variables products but not if it's an add to cart button as that's handled separately.

assets/js/src/integrations/classic.js
martynmjones: The first `removeFromCartListener()` call is to attach the event listener when the page first loads. If an item is then removed from the cart, WooCommerce replaces the cart div and dispatches an `updated_wc_div` event.

When the cart is replaced the initial event listener is destroyed which is why we need to use that event to attach a new one. So it won't track the same action twice but we need it to make sure we track every item removed from the cart.

assets/js/src/integrations/blocks.js
martynmjones: Thanks! Removed in bed5877

assets/js/src/tracker/index.js
martynmjones: Updated in 7fcab41

assets/js/src/tracker/index.js
martynmjones: Nice suggestion - The object was left over from a much earlier structure where we would have returned other functionality but you're right it's no longer needed so I've tested and merged your branch into this one. Thanks!

assets/js/src/integrations/classic.js
martynmjones: This is a little bit hacky but in our [event data](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/155b32f44da7e837419983cc0d24df2c90fca4ad/assets/js/src/integrations/classic.js#L22) we use `product` for single products. With the add to cart event that's dispatched by blocks we receive an object `{ product: ... }` so this was added to pass the data along to the formatter in the same way.

includes/class-wc-google-gtag-js.php
martynmjones: This was done so that when we're hooking into WooCommerce and adding product data for something like `view_item_list` we wouldn't queue up multiple of the same event. It definitely isn't particularly clear to read though so instead I've removed it and [added a separate check](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d6259842a405e9387b537a38a60c54ec48ddb168/includes/class-wc-google-gtag-js.php#L104).

includes/class-wc-google-gtag-js.php
martynmjones: Thanks for highlighting that @tomalec. It evolved over time and you're right it was structured in a fairly confusing way. I've instead simplified how it works and added a separate function to manage adding more data in c4931e3 (_Related to the [comment above](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328#discussion_r1484587515)_).

includes/class-wc-abstract-google-analytics-js.php
martynmjones: Updated in 46dd222

assets/js/src/config.js
martynmjones: Good point, I've added the [suggested changes](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d6259842a405e9387b537a38a60c54ec48ddb168/assets/js/src/config.js).

martynmjones:
Hey @tomalec, many thanks for your review and suggestions. I've made changes and responded to the questions you raised so this is ready for another review.

tomalec:
When I try to render the cart page, it throws an error:
```
nginx_1      | 2024/02/19 15:46:38 [error] 45#45: PHP Fatal error:  Uncaught Error: Cannot unpack array with string keys in /var/www/html/wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php:174
nginx_1      | Stack trace:
nginx_1      | #0 [internal function]: WC_Abstract_Google_Analytics_JS->{closure}()
nginx_1      | #1 /var/www/html/wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php(178): array_map()
nginx_1      | #2 /var/www/html/wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php(68): WC_Abstract_Google_Analytics_JS->get_formatted_cart()
nginx_1      | #3 /var/www/html/wp-includes/class-wp-hook.php(324): WC_Abstract_Google_Analytics_JS->{closure}()
nginx_1      | #4 /var/www/html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
nginx_1      | #5 /var/www/html/wp-includes/plugin.php(517): WP_Hook->do_action()
nginx_1      | #6 /var/www/html/wp-content/plugins/woocommerce/templates/cart/cart.php(20): do_action()
```

Did we drop PHP 7.x support?

martynmjones:
> When I try to render the cart page, it throws an error:
> 
> ```
> nginx_1      | 2024/02/19 15:46:38 [error] 45#45: PHP Fatal error:  Uncaught Error: Cannot unpack array with string keys in /var/www/html/wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php:174
> nginx_1      | Stack trace:
> nginx_1      | #0 [internal function]: WC_Abstract_Google_Analytics_JS->{closure}()
> nginx_1      | #1 /var/www/html/wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php(178): array_map()
> nginx_1      | #2 /var/www/html/wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php(68): WC_Abstract_Google_Analytics_JS->get_formatted_cart()
> nginx_1      | #3 /var/www/html/wp-includes/class-wp-hook.php(324): WC_Abstract_Google_Analytics_JS->{closure}()
> nginx_1      | #4 /var/www/html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
> nginx_1      | #5 /var/www/html/wp-includes/plugin.php(517): WP_Hook->do_action()
> nginx_1      | #6 /var/www/html/wp-content/plugins/woocommerce/templates/cart/cart.php(20): do_action()
> ```
> 
> Did we drop PHP 7.x support?

Apologies @tomalec, I've pushed an updated so you should no longer get that error 9f60881.

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,106 @@
...
+		document
+			.querySelector( '.woocommerce-cart-form' )
+			?.addEventListener( 'click', ( e ) => {
+				const item = e.target.closest(
+					'.woocommerce-cart-form__cart-item .remove'
+				);
+
+				if ( ! item || ! item.dataset.product_id ) {
+					return;
+				}
```
tomalec, line 66: :question: 
Technically, we do not _"Attach click event listeners to all remove from cart links"_ we do "Attach a click event listener to the entire form".
And I wonder why? Why do we listen to all the clicks anywhere in the form, instead of 

```js
document
    .querySelectorAll( '.woocommerce-cart-form .woocommerce-cart-form__cart-item .remove[data-product_id]')
    .forEach( (item) => item.addEventListener( 'click', removeFromCartHandler ) );
    
function removeFromCartHandler( e ) {
	tracker.eventHandler( 'remove_from_cart' )( {
		product: getProductFromID(
			parseInt( e.target.dataset.product_id ),
			true
		),
	} );
}    
```
Also, having the handler as the same, not anonymous function for all `addEventListener`s, would guard us against attaching the same listener twice. 
(as mentioned in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328#discussion_r1473645572)

----

The suggested code would not work if the remove from cart buttons are added asynchronously after the page load and `onupdated_wc_div`

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,106 @@
...
+
+				tracker.eventHandler( 'remove_from_cart' )( {
+					product: getProductFromID(
+						parseInt( item.dataset.product_id ),
+						true
+					),
+				} );
+			} );
```
tomalec: :question: 
Technically, we do not _"Attach click event listeners to all remove from cart links"_ we do "Attach a click event listener to the entire form".
And I wonder why? Why do we listen to all the clicks anywhere in the form, instead of 
```js
document
    .querySelectorAll( '.woocommerce-cart-form .woocommerce-cart-form__cart-item .remove[data-product_id]')
    .forEach( (item) => item.addEventListener( 'click', removeFromCartHandler ) );
    
function removeFromCartHandler( e ) {
	tracker.eventHandler( 'remove_from_cart' )( {
		product: getProductFromID(
			parseInt( e.target.dataset.product_id ),
			true
		),
	} );
} 
```

Also, having the handler as the same, not anonymous function for all `addEventListener`s, would guard us against attaching the same listener twice. 
(as mentioned in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328#discussion_r1473645572)

----

The suggested code would not work if the remove from cart buttons are added asynchronously after the page load and `onupdated_wc_div`

---


```suggestion
	/**
	 * Attach click event listeners to all remove from cart links on page load and when the cart is updated.
	 */
	const removeFromCartListener = () => {
		document
			.querySelectorAll(
				'.woocommerce-cart-form .woocommerce-cart-form__cart-item .remove[data-product_id]'
			)
			.forEach( ( item ) =>
				item.addEventListener( 'click', removeFromCartHandler )
			);
	}

	function removeFromCartHandler( e ) {
		tracker.eventHandler( 'remove_from_cart' )( {
			product: getProductFromID(
				parseInt( e.target.dataset.product_id ),
				true
			),
		} );
	}
```

assets/js/src/integrations/classic.js
tomalec: Thanks for the explanation.

So, the logic relies on the fact that WC will not call `updated_wc_div` before/without removing it.

I'm not sure how safe that assumption is. I could imagine one could understand that `updated_wc_div` could be called even if the div stays on the page, but a piece of its content gets replaced.

To be safe, I'd attach the same non-anonymous function as the event listener (see https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328/files#r1494819398)

assets/js/src/integrations/classic.js
tomalec: Sorry, I still don't get the logic.

- The current code does not guarantee that `button.parentNode` is a `.product`

- Also, for me clicking on the enclosing element (see below) does trigger the `select_content` event, even though it does not click on any link, or redirect anywhere
 
image

- What does `woocommerce-loop-product__link` has to do with the play, and when it appears? 



> We need to add this check to make sure the event does fire if someone clicks on the main button when it's for variables products but not if it's an add to cart button as that's handled separately.

Does the "main button" mean this:
image 
image

Then, if we want to capture "clicking anywhere a tile that contains those buttons", then I'd rather use
```js
	document
		.querySelectorAll(
			'.product a[data-product_id].product_type_variable, .product a[data-product_id]:not(.add_to_cart_button)'
		)
		?.forEach( ( selectContentButton ) => {
			const productId = selectContentButton.dataset.product_id;
			const productTile = selectContentButton.closest( '.product' );

			productTile.addEventListener( 'click', () => {
				tracker.eventHandler( 'select_content' )( {
					product: getProductFromID( parseInt( productId ) ),
				} );
			} );
		} );
```

assets/js/src/integrations/blocks.js
tomalec: Could you add a comment about why we remove those actions and who creates them?

includes/class-wc-google-gtag-js.php
```diff
@@ -37,498 +36,192 @@ public static function get_instance( $options = array() ) {
...
-		);
-
-		$select_content_event_code = self::get_event_code(
-			'select_content',
+	public function append_script_data( string $type, $data ): void {
+		$this->set_script_data(
+			$type,
 			array(
-				'items' => array( $item ),
+				...$this->script_data[ $type ] ?? [],
```
tomalec: :nail_care: I wonder if the performance of destructing and re-constructing a new array over and over, as we push products/events in the loop matters here. Maybe we can use something even shorter

```php
	public function append_script_data( string $type, $data ): void {
		if ( ! isset( $this->script_data[ $type ] ) ) {
			$this->script_data[ $type ] = array();
		}
		$this->script_data[ $type ][] = $data;
	}
```

tomalec (changes requested):
Thank you for all the changes and clarifications.

I left some comments. Many are cosmetic ones. The most concerning to me are:
- `select_content` logic https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328#discussion_r1473642148
- `remove_from_cart` handlers https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328#discussion_r1494819398

Also, while testing, I do not get `remove_from_cart` event triggered when I remove a variable item from the cart.
[Screencast from 19.02.2024 19:21:07.webm](https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/17435/81cb36e5-48e8-4504-a4db-4322321eb272)

The `trackerEventHandler` is called, but fails with
```
index.js:15 Uncaught TypeError: Cannot read properties of undefined (reading 'variation')
    at getProductFieldObject (index.js:15:15)
    at remove_from_cart (data-formatting.js:62:33)
    at trackerEventHandler (index.js:72:5)
    at HTMLFormElement.<anonymous> (classic.js:68:47)
```

tomalec:
Seems there is an id mismatch 
image

assets/js/src/integrations/classic.js
martynmjones: Updated in 9a1e058. I've restructured it slightly different to how we discussed earlier so it should be easier to read now.

assets/js/src/integrations/classic.js
martynmjones: Thanks for the suggestions @tomalec. Updated in 234fe84.

assets/js/src/integrations/blocks.js
martynmjones: Comment added 8272538

includes/class-wc-google-gtag-js.php
martynmjones: Updated in c979cf0

martynmjones:
The changes we discussed earlier have been made and the ID mismatch has been resolved so this is ready for another round @tomalec. Thanks!

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,132 @@
...
+				.querySelector( 'a[data-product_id]' )
+				?.getAttribute( 'data-product_id' );
+
+			if ( ! productId ) {
+				return;
+			}
+
+			item.addEventListener( 'click', ( event ) => {
+				// Return early if the user has clicked on anything other
+				// than a product link or an Add to cart button.
```
tomalec: :nail_care: Sorry, but as a non-native English speaker I read it as "the user has clicked on anything other than (a product link or an Add to cart button)" and I think you mean "the user has clicked on (anything other than a product link) or an Add to cart button", maybe something like this would be safer:

```suggestion
				// Return early if the user has clicked on 
				// an "Add to cart" button or anything other than a product link
```

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,132 @@
...
+
+				const isAddToCartButton =
+					event.target.classList.contains( 'add_to_cart_button' ) &&
+					! event.target.classList.contains(
+						'product_type_variable'
+					);
+
+				if ( ! targetLink && ( ! isButton || isAddToCartButton ) ) {
+					return;
+				}
```
tomalec: :nail_care: 
I got a gut feeling that this selector may be hard to keep in sync to match with WC core/theme changes.

Should we fire the `add_to_cart` event for non-ajax buttons as well?
From reading WC Core code, I think we can disable `ajax_add_to_cart` globally or for a product, then end up with a 
`<a class="button add_to_cart_button">Add to cart</a>`.
If we want to fire for those, then the `!isButton` condition should be removed. Otherwise, we could use `.ajax_add_to_cart` to precisely select only those add-to-cart-like buttons that actually add to cart, guarding against `.product_type_variable` to any other.

Are you sure we will not have some other `.button` in the product card that does not actually select the product, but does some other thing? Like a link to a promotion, group, brand site, etc?
Maybe we can identify "Add to cart"-likeness by having `data-product_id` attribute :) 
So we would do 
```suggestion
				const isProductLink =
					// Regular product link.
					event.target.closest( '.woocommerce-loop-product__link' ) ||
					// Add-to-cart-like button.
					event.target.hasAttribute( 'data-product_id' );

				const isAddToCartButton =
					event.target.classList.contains( 'add_to_cart_button' ) &&
					! event.target.classList.contains(
						'product_type_variable'
					);

				if ( ! isProductLink || isAddToCartButton ) {
					return;
				}
```


Also, all this confusion looks like an encouragement to champion a PR to WC Core to add the product_id datar attribute to the product card ;)

assets/js/src/integrations/classic.js
```diff
@@ -0,0 +1,132 @@
...
+	document.body.onupdated_wc_div = () => removeFromCartListener();
+
+	// Trigger the handler when an item is removed from the mini-cart and WooCommerce dispatches the `removed_from_cart` event.
+	document.body.onremoved_from_cart = (
+		event,
+		fragments,
+		/* eslint-disable-next-line camelcase */
+		cart_hash,
+		button
+	) => removeFromCartHandler( { target: button[ 0 ] } );
```
tomalec: As we are not using `on` or `addEventListener`, but hard-setting the property, we could be more gentle towards other plugins, trying to hook into those events:

```suggestion
	const oldOnupdatedWcDiv = document.body.onupdated_wc_div;
	document.body.onupdated_wc_div = ( ...args ) => {
		if ( typeof oldOnupdatedWcDiv === 'function' ) {
			oldOnupdatedWcDiv( ...args );
		}
		removeFromCartListener();
	};

	// Trigger the handler when an item is removed from the mini-cart and WooCommerce dispatches the `removed_from_cart` event.
	const oldOnRemovedFromCart = document.body.onremoved_from_cart;
	document.body.onremoved_from_cart = ( ...args ) => {
		if ( typeof oldOnRemovedFromCart === 'function' ) {
			oldOnRemovedFromCart( ...args );
		}
		removeFromCartHandler( { target: args[ 3 ][ 0 ] } );
	};
```

tomalec (approved):
Thank you for the changes and explanations. 
Left a few tweak ideas
LGTM.

assets/js/src/integrations/classic.js
martynmjones: Thanks for highlighting that! Updated in 94cc036

assets/js/src/integrations/classic.js
martynmjones: When clicking on non-ajax add to cart button the event will be sent when the page [loads and the confirmation message is displayed](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/94cc036583bc03f5b9ef6dd829401ab9a4207225/includes/class-wc-abstract-google-analytics-js.php#L80) so I don't think we need additional targeting there.

I've added a check like you suggested so that the event will only send if the button includes the `data-product_id` attribute to reduce the chance of issues with other extensions in 94cc036.

PR #335, opened 2023-11-24
Update test environment setup to resolve notice

martynmjones:
### Changes proposed in this Pull Request:

Closes #334 

Updates `UnitTestsBootstrap` to run `install_wc` on the `init` hook to resolve the notice.

### Screenshots:
Screenshot 2023-11-24 at 15 00 07

### Detailed test instructions:
1. Confirm tests pass without triggering a notice using the latest version of WooCommerce

### Additional

- PHP unit tests failed but they are fixed here: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/333. Local tests passes if you run `composer require --dev yoast/phpunit-polyfills:"^1.1.0" && composer install`

### Changelog entry

> Tweak - Test environment setup to resolve notice

PR #339, opened 2023-12-15
Prevent DB warnings in unit tests

### LINKED ISSUE #338: Unit testing outputs warnings when starting with a clean database
### Describe the bug:
We recently changed the order in bootstrapping the WooCommerce extension, see #335 

However the consequence of that change is it doesn't recreate the DB tables at the right time. So when we run the unit tests with a clean database it complains that some tables are non existent.
See a recent run: https://github.com/woocommerce/woocommerce-google-analytics-integration/actions/runs/7022872556/job/19108120718
image

I think we should revert #335 and instead filter the notice like we do in other extensions, so the bootstrapping of WooCommerce can be done earlier again, allowing it to recreate the required tables when installing during unit tests. See: https://github.com/woocommerce/automatewoo/pull/1642

### Steps to reproduce:
1. Install unit tests with a version of WC > 8.3 `bin/install-unit-tests.sh <db_name> <db_user> <db_pass>` (select `Y` to clear database during install)
2. Run unit tests `vendor/bin/phpunit`
3. Notice DB warnings before tests are run

### Expected behavior:
No warnings when running unit tests.

mikkamp:

This PR reverts the change in #335 since it still produces warnings when starting tests with an empty database. So here we revert to an earlier hook and add a filter to prevent the notice from showing. We also use `WC_REMOVE_ALL_DATA` so all database tables can be removed and recreated when `WC_Install` is called.

The PR also adds PHP version 8.3 to the workflow, so it now runs the default tests on PHP 8.2 and includes the oldest supported version (7.4) and the latest supported version (8.3).

Closes #338

1. Install unit tests with a version of WC > 8.3 `bin/install-unit-tests.sh <db_name> <db_user> <db_pass>` (select `Y` to clear database during install)
2. Run unit tests `vendor/bin/phpunit`
3. Confirm there are no longer any warnings when the tests are run
4. Check the Workflow tests and confirm we have successful runs for PHP 7.4, 8.2, 8.3

### Changelog entry
* Dev - Prevent DB warnings in unit tests.

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -45,7 +45,7 @@ jobs:
     needs: GetMatrix
     strategy:
       matrix:
-        php: [8.1]
+        php: [8.2]
         wp-version: [latest]
         # Please note that wc-versions is a string containing versions separated by commas.
         # It will be split and loop within the run unit test step below to reduce the time spent.
@@ -54,7 +54,7 @@ jobs:
           - php: 7.4
             wp-version: ${{ fromJson(needs.GetMatrix.outputs.wp-versions)[2] }} # L-2 WP Version support
             wc-versions: ${{ fromJson(needs.GetMatrix.outputs.wc-versions)[2] }} # L-2 WC Version support
-          - php: 8.2
+          - php: 8.3
             wp-version: latest
             wc-versions: latest
 
--- a/tests/class-unittestsbootstrap.php
+++ b/tests/class-unittestsbootstrap.php
@@ -102,33 +102,15 @@ class UnitTestsBootstrap {
 		echo 'Installing WooCommerce...' . PHP_EOL;
 
 		define( 'WP_UNINSTALL_PLUGIN', true );
+		define( 'WC_REMOVE_ALL_DATA', true );
 
 		include $this->plugins_dir . '/woocommerce/uninstall.php';
 
-		global $wpdb;
-		$wpdb->query( "DROP TABLE IF EXISTS {$wpdb->prefix}woocommerce_attribute_taxonomies" );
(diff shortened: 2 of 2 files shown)


puntope (approved):
Thanks @mikkamp  I don't see anymore the errors in the console and the tests pass. Also now the file seems consistent with the AW one.

❓  In https://github.com/woocommerce/google-listings-and-ads/pull/2183 you mentioned that we should mark extensions as compatible with PHP 8.3, Should we mark this one as Composer as well as the tests run smoothly? Or it's any other step to take?

mikkamp:
> ❓In https://github.com/woocommerce/google-listings-and-ads/pull/2183 you mentioned that we should mark extensions as compatible with PHP 8.3, Should we mark this one as Composer as well as the tests run smoothly? Or it's any other step to take?

I've been marking them as compatible in the project thread: peeuvX-19l-p2
I'm not sure I follow about marking it as compatible in composer, I don't recall us generally setting any max PHP version, at most we have extensions which set a minimum PHP version: https://github.com/woocommerce/google-listings-and-ads/blob/2.5.13/composer.json#L8
But that's only for extensions that include composer packages in the release.

puntope:
> > ❓In https://github.com/woocommerce/google-listings-and-ads/pull/2183 you mentioned that we should mark extensions as compatible with PHP 8.3, Should we mark this one as Composer as well as the tests run smoothly? Or it's any other step to take?
> 
> I've been marking them as compatible in the project thread: peeuvX-19l-p2 I'm not sure I follow about marking it as compatible in composer, I don't recall us generally setting any max PHP version, at most we have extensions which set a minimum PHP version: https://github.com/woocommerce/google-listings-and-ads/blob/2.5.13/composer.json#L8 But that's only for extensions that include composer packages in the release.


What I meant about composer is that Composer install runs successfully. As far as I read that's a condition to make it compatible.

PR #353, opened 2024-01-31
Update plugin branding to Google Analytics for WooCommerce

martynmjones:

Renames the extension from `WooCommerce Google Analytics Integration` to `Google Analytics for WooCommerce`.

1. Review changes and confirm the extension name has been replaced correctly
2. Search codebase to confirm that there are no remaining instances of `WooCommerce Google Analytics Integration` that need to be replaced

### Additional details:

The plugin name in the main plugin file header comment has not been altered so as to not cause problems on WordPress.org.

### Changelog entry

> Update - Extension branding to Google Analytics for WooCommerce

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,14 +1,14 @@
 # Contributing
 
-Thanks for your interest in contributing to WooCommerce Google Analytics Integration!
+Thanks for your interest in contributing to Google Analytics for WooCommerce!
 
 ## Guidelines
 
 Like the WooCommerce project, we want to ensure a welcoming environment for everyone. With that in mind, all contributors are expected to follow our [Code of Conduct](./CODE_OF_CONDUCT.md).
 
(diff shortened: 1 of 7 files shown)


mikkamp (commented):
Thanks for the changes, they look good to me.

It seems there are still a few other places where it would be nice to match the new name.

- package.json title
- woocommerce-google-analytics-integration.php Plugin Name
- function woocommerce_missing_notice in both messages
- WC_Google_Analytics_Task::get_title
- Generated .pot file - Project-Id-Version (seems to depend on Plugin Name in main plugin file)

martynmjones:
Thanks for the review @mikkamp 

I'd misunderstood the guidance on .org for the `Plugin Name` header but you're right that can be replaced. I've updated that along with the package.json title, and notices thanks for catching that. The `.pot` file is updated when the extension is built so no changes needed there after updating the plugin name.

You also mentioned `WC_Google_Analytics_Task::get_title` which at the moment is `Set up Google Analytics` and doesn't reference the extension name. Do you think that needs updating too?

mikkamp (approved):
Thanks for the additional changes. Those look good.

> You also mentioned WC_Google_Analytics_Task::get_title which at the moment is Set up Google Analytics and doesn't reference the extension name. Do you think that needs updating too?

I just saw it is a task in the WC onboarding to prompt them to setup the extension. For GLA we have "Set up Google Listings & Ads". But since the link points to the integration settings which is called just "Google Analytics", I think either one would work. Could there also be another plugin that would use the same title? Probably a small chance but possible in theory. So I'll let you decide what you think is best.

Issue #359, opened 2024-02-23
`ga_ecommerce_tracking_enabled` is not sent to tracker data


tomalec:
### Describe the bug:

In [the remove-universal-analytics branch](https://github.com/woocommerce/woocommerce-google-analytics-integration/tree/remove-universal-analytics) `ga_ecommerce_tracking_enabled` setting is not sent to tracker data.

### Steps to reproduce:

1.  Go to _WooCommerce > Settings > Integration > Google Analytics_ `wp-admin/admin.php?page=wc-settings&tab=integration&section=google_analytics`
2. Make sure _" Purchase Transactions"_ is checked
3. Check the value of `ga_ecommerce_tracking_enabled` in  tracker data 
   ```
   wp wc tracker snapshot --format=json | jq '.[0]["wc-google-analytics"].ecommerce_tracking_enabled' 
   ```

### Expected behavior:

`"yes"` should be tracked

### Actual behavior:

`null` is tracked

### Additional details:

1. I also wonder if WCGAI stores its data in the correct path. Other extensions (GLA, AW) add its data under the `.extensions` namespace, like 
   ```
   wp wc tracker snapshot --format=json | jq ".[0].extensions.gla.settings.version"
   ```

### Investigation notes
It seems our `options` logic that to rewrite/duplicate `->settings` misses the `ga_ecommerce_tracking_enabled` key
 - `->settings` has it
 - `->form_fields` has it
 - `track_options` expects it to be in `$this`
 - `$this->` has a property to store it
 - :bug: `init_options` misses it in rewriting from `getOptions` to `$this->`

tomalec:
A quick fix would be to add `ga_ecommerce_tracking_enabled` to the mapping list at https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/9f60881ef67756acf9ef514100e11244f1e3fc63/includes/class-wc-google-analytics.php#L145-L159
But I wonder if we actually need to map it back and forth, risking the duplicated list going out of sync again, instead of using `->settings` directly.

tomalec:
Also, I noted that the `ga_id` is missing in the tracking report

PR #360, opened 2024-02-26
Remove options remapping, use settings directly.

tomalec:
_:warning: This PR requires https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/328_

- [Remove options remapping, use settings directly.](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/0bc1a5b715cfb1debc264976a14b2a5a03738c22) 
   Fix https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/359

I don't see any value in having another layer that's just a copy of the existing `->settings`. Having three places to store the same thing may lead to more bugs like #359, where the objects get out of sync.
- [Remove redundant WC_Google_Analytics properties](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/899c0d3d0561487b78b57174b022106ea32f8316) 
   use `$this->settings` directly.
   All the properties, except `ga_id`, were not used at all. `ga_id` was used only once.
   
### Checks:

1.  Go to _WooCommerce > Settings > Integration > Google Analytics_ `wp-admin/admin.php?page=wc-settings&tab=integration&section=google_analytics`
2. Make sure _" Purchase Transactions"_ is checked
3. Check the value of all the settings match the tracker data 
   ```
   wp wc tracker snapshot --format=json | jq '.[0]["wc-google-analytics"]' 
   ```
4. Check that `wcgaiData.config` matches the settings
5. Smoke test the plugin

### Additional details:
1. I think we can consider one more tweak: removing the `static get` function and using the settings array directly. A static function that relies on the property set by the instance constructor looks bug-prone. Plus, defining a function to do `$this->get( 'foo' )` instead of `$this->settings['foo']`, makes the code less explicit and does not bring much value IMO

### Changelog entry

> Dev - Remove options remapping, use settings directly.

### DIFF
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -17,8 +17,8 @@ abstract class WC_Abstract_Google_Analytics_JS {
 	/** @var WC_Abstract_Google_Analytics_JS $instance Class Instance */
 	protected static $instance;
 
-	/** @var array $options Inherited Analytics options */
-	protected static $options;
+	/** @var array $settings Inherited Analytics settings */
+	protected static $settings;
 
 	/** @var string Developer ID */
 	public const DEVELOPER_ID = 'dOGY3NW';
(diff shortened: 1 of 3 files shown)


martynmjones (approved):
Hey @tomalec, thanks for tidying up the extension settings 🙌 

Tracking is working as expected and the correct events are being loaded into the `wcgaiData` object so LGTM ✅

PR #361, opened 2024-02-27
Implement consent mode v0

tomalec:

Add the minimalistic implementation to add support for [Google Consent mode](https://developers.google.com/tag-platform/security/guides/consent?consentmode=advanced)

We want to add MVP ASAP before Google starts penalizing merchants without this setup. (pcTzPl-25o-p2)
Later, we could iterate by adding settings to customize the snippet a bit using UI.

This PR adds a basic default and documents how to customize/extend it.

### Checks:

### Screenshots:
image

1. Buiild, Install and activate the extension
2. Deactivate GLA (if applicable, as the integration is broken for `remove-universal-analytics` branch)
3. Disable "Add to cart" events in settings (`/wp-admin/admin.php?page=wc-settings&tab=integration&section=google_analytics`) so WooCommerce Core will not set up its own `gtag` config making us do nothing.
4. Open [Tag Assistant](https://tagassistant.google.com/) 
5. Open shop page the page 
6. In Tag assistant you should see the "Consent" as the first entry.
7. If you're connecting from EEA region, you should see "Consent" being Denied by default
   image
8. If you're connecting from non-EEA region, you should see "Consent" not set 
   image
10. Check that Readme description
11. Add the snippet to customize the consent mode:
```php
add_action( 'wp_enqueue_scripts', function () {
    $code = "
        gtag( 'consent', 'default', {
            analytics_storage: 'granted',
            ad_storage: 'granted',
            ad_user_data: 'denied',
            ad_personalization: 'granted',
            region: 'PL', // <--------- SET IT TO YOUR REGION
        } );";

    wp_register_script( 'my-custom-consent-mode', '', array('woocommerce-google-analytics-integration'), null, [ 'in_footer' => true ] );
    wp_add_inline_script( 'my-custom-consent-mode', $code );
    wp_enqueue_script( 'my-custom-consent-mode' );
} );
```
Check that the second Consent entry is added with custom defaults
image

### Additional details:
1. This customization capabilities are not ideal, as the is a change to get `gtag` hit before the second consent config is saved, as we may trigger page_view or other events. We may improve that in the future. 

   Especially since the current logic to force `main.js` to run after the document is ready is pretty hacky. (I guess that's why we add inline script` on `wp_footer` hook). I think we can make it in a cleaner fashion:
   - wrap all JS code that requires document ready in `DOMContentLoaded` event listener
   - execute the code that needs to run synchronously

### Changelog entry

> Add - Basic consent mode support.


tomalec:
After https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/363 the `[ 'in_footer' => true ]` part would not be needed in the snippet.

assets/js/src/config.js
```diff
@@ -1,2 +1,36 @@
...
+	'NL',
+	'NO',
+	'PL',
+	'PT',
+	'RO',
+	'SK',
+	'SI',
+	'ES',
+	'SE',
+	'UK',
```
martynmjones: I know this is based on the snippet Google provided but `UK` doesn't appear to be a valid country code. Testing the branch as it, there is no `On-page default` set:

Screenshot 2024-03-01 at 12 46 17

Updating it to `GB` fixes that:

Screenshot 2024-03-01 at 12 46 44

```suggestion
	'GB',
```

martynmjones (approved):
Hey @tomalec, thanks for adding the initial consent implementation.

All appears to be working well and the consent parameters are picked up by Google. One issue with the country code config but everything else looks good ✅

PR #362, opened 2024-02-28
Restructure gtag and event data config loading

martynmjones:

Closes # .

_Replace this with a good description of your changes & reasoning._

### Checks:

1. 
2. 
3. 

### Changelog entry

>

### DIFF
--- a/assets/js/src/config.js
+++ b/assets/js/src/config.js
@@ -1,11 +1,7 @@
 /* global wcgaiData */
 /* eslint-disable camelcase */
-export const {
-	config,
-	events,
-	cart,
-	products,
-	product,
-	added_to_cart: addedToCart,
-	order,
-} = wcgaiData;
+export const eventData = function() {
+	return wcgaiData;
+};
+
+export const trackerData = wcgaiTracker;
--- a/assets/js/src/config.js
+++ b/assets/js/src/config.js
@@ -1,6 +1,6 @@
 /* global wcgaiData */
-/* eslint-disable camelcase */
-export const eventData = function() {
+/* global wcgaiTracker */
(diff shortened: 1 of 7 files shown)


assets/js/src/config.js
```diff
@@ -1,11 +1,7 @@
...
-	config,
-	events,
-	cart,
-	products,
-	product,
-	added_to_cart: addedToCart,
-	order,
-} = wcgaiData;
+export const eventData = function() {
+	return wcgaiData;
```
tomalec, line 4: What's the point fo having this as a function? instead of returning it directly?

assets/js/src/index.js
```diff
@@ -1,6 +1,12 @@
+import { eventData } from './config.js';
+import { trackClassicIntegration } from './integrations/classic'
+
 // Initialize tracking for classic WooCommerce pages
-import { trackClassicIntegration } from './integrations/classic';
-trackClassicIntegration();
+document.addEventListener( 'DOMContentLoaded', () => {
```
tomalec, line 5: To be on the safe side we'd need to cater for the case when the event already loaded https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/legacy/js/frontend/order-attribution.js#L123-L127 this way we can include the script before and after and it will do the job.

assets/js/src/config.js
martynmjones: `wcgaiData` won't necessarily be defined when this file loads so if we return it directly then we'll get an error and it won't be usable.

But because `eventData()` is only used by the classic integration which runs on the `DOMContentLoaded` event it will be accessible in the footer.

PR #364, opened 2024-03-04
Allow more precise consent mode customizations

tomalec:

Allow more precise consent mode customizations. Use the filtered array with potential customizations just after `gtag` is defied, so the consent mode is set up before it's used, allowing cookies to be created correctly for recurring `granted` visitors.

### Checks:

1. Build, Install and activate the extension
4. Open [Tag Assistant](https://tagassistant.google.com/) 
5. Open shop page the page 
6. In Tag assistant you should see the "Consent" as the first entry.
7. If you're connecting from EEA region, you should see "Consent" being Denied by default
   image
8. If you're connecting from non-EEA region, you should see "Consent" not set 
   image
10. Check the Readme description 
11. Add the snippet to customize the consent mode:
```php
add_filter( 'woocommerce_ga_gtag_consent_modes', function ( $consent_modes ) {
    $consent_modes[] =
		array(
            'analytics_storage' => 'granted',
            'region'            => array( 'PL' ), // <------ your region
        );
   return $consent_modes;
} );
```
12. Check that the second Consent entry is added with custom defaults before the container is loaded
   image
13. Check that the cookies were created
   ```js
   (await cookieStore.getAll()).filter((c)=>c.name.includes('ga'))
   ```
   image

### Changelog entry
_no changelog entry, as this is a part of the new to-be-released feature_


martynmjones (approved):
Hey @tomalec, thanks for quickly fixing the issue with consent defaults 🙌 

Adding region defaults via filter works well and in testing I was seeing the correct cookies being set at the expected time so users are properly tracked across multiple page loads. ✅

PR #322, opened 2023-10-30
Remove Universal Analytics

martynmjones:
### Changes proposed in this Pull Request:

Closes #215

- [x] #309
- [x] #320
- [x] #311
- [x] #321
- [x] #328

### Checks:
* [x] Does your code follow the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards?
* [x] Have you written new tests for your changes, as applicable?
* [x] Have you successfully run tests with your changes locally?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changelog entry

> Remove - Universal Analytics integration

PR #365, opened 2024-03-05
Prevent tracking admins or order multiple times and additional adjustments

martynmjones:

Following on from #322, this PR ensures that no tracking happens if [disable_tracking()](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/0a546f42430ad9fcda6f9b17a3840a9acb9a3bc2/includes/class-wc-google-analytics.php#L43) returns true. Additionally, the new tracking wasn't considering the `_ga_tracked` meta value for orders meaning if a user was to refresh the thank you page multiple `purchase` events would be triggered which is fixed in 94b4dc7.

**Additional cleanup:**

- Removed `async_script_loader_tags()` and set `strategy` on the enqueued script 4e8e1ac
- Added filter to [adjust gtag config](https://github.com/woocommerce/woocommerce-google-analytics-integration/commit/e5485bf1ddc6a719a8e6c0b7d2730989dbc0356e)
- Added check to confirm specific event is enabled before tracking d7befb1 (_This was added earlier but lost in a previous PR_)
- Removed settings that are no longer relevant with GA4 3b26491
    1. `Enable Standard Tracking`
        - Previously had an impact on how things were loaded but didn't do much beyond that. The option description did not seem to be particularly relevant either so I thought the best option was to remove it entirely
    2. `Use Enhanced Link Attribution`
        - No longer relevant. See https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-link-attribution
    4. `Anonymize IP addresses`
       - No longer relevant. See https://support.google.com/analytics/answer/2763052

### Screenshots:

Screenshot 2024-03-05 at 13 00 08

1. Build extension from `fix/tracking-for-purchases-and-admins`
2. Go to [Tag Assistant](https://tagassistant.google.com/) and debug site
3. While logged in as an admin, confirm that no tracking is sent
4. Logout
5. Open an incognito window and ensure all options are selected in `WooCommerce > Settings > Integrations > Google Analytics`
6. Confirm events are tracked
7. Place an order and confirm a `purchase` event is sent
8. Refresh the thank you page and confirm a second `purchase` event is not sent
9. In the incognito window, disable some events like `Add to Cart Events` and `Product Detail Views`
10. In the non-incognito window, confirm that those events are no longer sent

### Additional:

I believe we can remove a couple of the other settings but as we're close to release I decided to only remove settings that explicitly state they are not support by GA4. If any other changes are needed that can be addressed post launch.

### Changelog entry

> Fix - Prevent tracking orders multiple times and additional fixes

### DIFF
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -173,6 +173,10 @@ export const view_item = ( {
  * @param {Object} params.order The order object
  */
 export const purchase = ( { order } ) => {
+	if ( order === undefined ) {
+		return false;
+	}
+
 	return {
 		currency: order.currency,
 		value: parseInt( order.value ),
--- a/assets/js/src/tracker/index.js
+++ b/assets/js/src/tracker/index.js
@@ -71,11 +71,13 @@ class Tracker {
 		}
 
 		return function trackerEventHandler( data ) {
-			window[ config.tracker_function_name ](
-				'event',
-				name,
-				formatter( data )
-			);
+			if ( config.events.includes( name ) ) {
+				window[ config.tracker_function_name ](
+					'event',
+					name,
+					formatter( data )
+				);
+			}
 		};
 	}
 }
--- a/assets/js/src/tracker/index.js
+++ b/assets/js/src/tracker/index.js
@@ -71,11 +71,12 @@ class Tracker {
 		}
 
 		return function trackerEventHandler( data ) {
-			if ( config.events.includes( name ) ) {
+			const eventData = formatter( data );
+			if ( config.events.includes( name ) && eventData ) {
 				window[ config.tracker_function_name ](
 					'event',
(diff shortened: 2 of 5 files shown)


tomalec (approved):
Thank you for fixes and cleanups!
Reviewed the code, tested locally, LGTM :)

The only comment I have is that as we remove `ga_standard_tracking_enabled` setting from here, we need to accommodate for that change in GLA integration.

martynmjones:
> Thank you for fixes and cleanups! Reviewed the code, tested locally, LGTM :)
> 
> The only comment I have is that as we remove `ga_standard_tracking_enabled` setting from here, we need to accommodate for that change in GLA integration.

Thanks @tomalec! The standard tracking reference in GLA is addressed in https://github.com/woocommerce/google-listings-and-ads/pull/2297

PR #370, opened 2024-03-07
Resolve Google Analytics not loading if gtag already exists on the window object.

### LINKED ISSUE #367: Plugin does not load tracker if gtag already exists on the window object
The plugin does not load the tracker if gtag already exists on the window object.

https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/70790675433c81ac83879cdf67fc36ddd238b5d0/assets/js/src/tracker/index.js#L29

This can happen if you're using another plugin or method to implement a different Google property such as Google Ads for example. This change breaks the plugin across all our clients completely disabling GA if they update to 2.0 😬

workeffortwaste:

Closes https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/367.

Resolves the tracker not loading if gtag already exists on the window object.

It's valid for gtag to already exist on the object, as a cookie banner may have created it to send a consent update, or it may exist from another tag such as Google Ads.

Here's an example from Google where it's valid for it to be created multiple times: 

https://developers.google.com/tag-platform/security/guides/consent?consentmode=advanced#gtag.js_3

### Checks:

1. Initialise a gtag object on the window before the plugin loads its JS.

Barebones example for testing purposes:
```
<script>
</script>
```
3. Confirm that the plugin has successfully loaded the Google Analytics tag.

### DIFF
--- a/assets/js/src/tracker/index.js
+++ b/assets/js/src/tracker/index.js
@@ -26,11 +26,6 @@ class Tracker {
 	 * Initializes the tracker and dataLayer if not already done.
 	 */
 	init() {
-		if ( window[ config.tracker_function_name ] ) {
-			// Tracker already initialized. Do nothing.
-			return;
-		}
-
 		window.dataLayer = window.dataLayer || [];
 
 		function gtag() {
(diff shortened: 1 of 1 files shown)


tomalec (approved):
Thank you for the fix!

I reviewed the code and tested it locally LGTM.

I've also removed the `init` function, moving its contents to the constructor. I see no reason to keep it publicly available.
With its contents in the constructor, we're sure we will not re-initialize the `gtag` ourselves

Issue #371, opened 2024-03-07
Critical errors with v.2.0.0


SirRahikkala:
When adding zero value products to cart the plugin gives Uncaught TypeError: array_keys(): Argument #1 ($array) must be of type array, string given in /wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-abstract-google-analytics-js.php:244 

nginx/1.24.0, PHP 8.0.30, Woocommerce 8.6.1,

tomalec:
Hi @SirRahikkala, We were not able to reproduce the issue with zero-value products.

But we managed to get the same error by mimicking (via code) a product with no attributes.
https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/372
Please let me know if that solves the problem for you.

If not, could you please give more detailed steps to reproduce your case?

SirRahikkala:
The products which see this error do have attributes which are weights so this is not about those. Could these screenshots point more forward?

![Näyttökuva 2024-03-07 193040](https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/7370720/0c30e4ae-0c62-4044-9ffc-e1e9e62bd4ff)


![Näyttökuva 2024-03-07 192923](https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/7370720/e775acac-5a39-4b4c-b3ac-5c35803f1b6e)

SirRahikkala:
Update 2.0.1 fixed the issue.

PR #372, opened 2024-03-07
Fix formatting product when variation has no attributes

tomalec:

quick-fix formatting product when variation has no attributes

Address https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/371

I was not able to reproduce it using UI, but according to https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/371 there is a chance to get `woocommerce_add_to_cart` fired for a product with `$variation` other than `false` but with no attributes.

### Checks:

1. ?
2. Mock it by adding `$variation = []` to https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/7594c1bf08355e5063b6a12acead52e594328feb/includes/class-wc-abstract-google-analytics-js.php#L232

### Changelog entry

> Fix - adding variable product attributes.

### DIFF
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -204,6 +204,7 @@ abstract class WC_Abstract_Google_Analytics_JS {
 	 *
 	 * @param WC_Product $product   The product to format.
 	 * @param array|bool $variation An array containing product variation attributes to include in the product data.
+	 *                              For the "variation" type products, we'll use product->get_attributes.
 	 *
 	 * @return array
 	 */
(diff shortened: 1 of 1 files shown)


includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -230,7 +230,7 @@ public function get_formatted_product( WC_Product $product, $variation = false )
 			$variation = $product->get_attributes();
 		}
 
-		if ( false !== $variation ) {
+		if ( false !== $variation && is_array( $variation ) ) {
```
jorgemd24: I think just having `if ( is_array( $variation ) )` would be sufficient, so we can remove the `false !== $variation`

tomalec: I removed the `variation` function param completely.

tomalec: Brought back the `$variation`, so we'd use it for  non-"variation" producs

jorgemd24 (approved):
Thanks! LGTM!

Issue #312, opened 2023-10-05
Issue with tracking quantity changes in cart


martynmjones:
### Describe the bug:
At the moment we're tracking changes to [product quantities in the cart](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/470dc91c0c02f2cb3816563be62005ce914ea453/assets/js/src/tracking.js#L82) (_when using blocks_) with a `change_cart_quantity` event. This isn't an event that's supported by [UA](https://developers.google.com/analytics/devguides/collection/gtagjs/events#default_google_analytics_events) or [GA4](https://developers.google.com/analytics/devguides/collection/ga4/reference/events).

It could be used as a custom event but additional reporting would need to be set up in Google Analytics for that and there's no reference to it in the documentation so I'm unsure of its original purpose.

From what I can gather, it appears that the correct way to register the events would be to send `add_to_cart` or `remove_from_cart` events with the number of items removed.

Unfortunately, the current (_experimental_) `-cart-set-item-quantity` action gives us the total number of a product in the cart but not the difference. We'll likely either need to open a PR in blocks to change that or rely on the cart state instead but it will require further exploration.

PR #375, opened 2024-03-07
Remove unused function from formatters

tomalec:

Remove three unused functions that seem to be leftovers from the old implementation:
- trackChangeCartItemQuantity
- trackException
- trackEvent

### Checks:

1. Double-check that none of those functions is used in the code
2. Smoke test

### Changelog entry

> Dev - Remove unused functions from code.


puntope (approved):
✅  LGTM

For the `trackChangeCartItemQuantity` seems like `change_cart_quantity` is not supported anymore by GA?

mikkamp:
Another one that doesn't seem to be used is `add_shipping_info`. Can we confirm if it was the intention to include this information during either `begin_checkout` or the `purchase` event?

puntope:
Re- Cart item quantity https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/312

tomalec:
@martynmjones Could you confirm regarding car quantities?

martynmjones:
> Another one that doesn't seem to be used is `add_shipping_info`. Can we confirm if it was the intention to include this information during either begin_checkout or the purchase event?

>  @martynmjones Could you confirm regarding car quantities?

We weren't previously tracking quantity changes or shipping info. Both those events had been added in the initial integration with Blocks in core but in testing neither were reliable and we didn't have an alternative for classic.

So as the extension was already not tracking those events we're safe to remove them. We can then look into alternatives at a later date if we decide we want to track the additional information.

tomalec:
@martynmjones, and what about [tracking shipping and `add_shipping_info`](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/375#issuecomment-1987887419)? 
Previously we were tracking it in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/322/files#diff-aa47bb8c923774f3c31519cb78bd1ae693f025fe39846d2571eb13640b399a87L64-L74

Should we add `checkout-set-shipping-address` to blocks.js?

martynmjones:
Sorry Tomek, I wasn't clear in my previous reply. The `add_shipping_info` event was only added in the core integration for Blocks and then copied over when we moved [tracking for Blocks to the extension](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/269).

Before that it didn't exist in the extension and there was never an equivalent for classic pages so I think we're safe to remove it.

PR #376, opened 2024-03-08
Remove unused fallback and undefined variable.

jorgemd24:

The below fallback is unnecessary as we always pass an Order object to the method  `get_formatted_order`, also there was a typo with the variable name `$order_id` as it should be `$order` and has been addressed while removing the fallback:

```
if ( is_int( $order ) ) {
	$order = wc_get_order( $order_id );
}
```

`get_formatted_order` is always called with an Order object.

https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/d8f43f6ecb58747d58219b0d454bae3464893b3d/includes/class-wc-abstract-google-analytics-js.php#L101-L113

### Checks:

1. Open an incognito window.
2. Place an order.
3. In the thank you page see if the tracking script has been rendered correctly in the source code:

image

### Changelog entry

> Fix - Typo with undefined variable.


mikkamp (approved):
Thanks for tweaking this code. It looks good to me, and I can confirm there are no cases where we call this function with a numeric order ID.

PR #378, opened 2024-03-08
Fix - Selectors in Product Blocks

### LINKED ISSUE #377: v2.0.0 - Some events doesn't trigger anymore
### Describe the bug:

After the update, even with all the consent granted, some of the track events are not being triggered properly.

### Affected events
 
- [ ] Shop Page - Add to cart
- [ ] shop page - view item List

### Expected behavior:

Same as in v1.8.0

https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/79f5fbd0-7399-4a7e-a663-377e64accec2

https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/d595b40f-7eb0-40ff-a7f4-5262590626d3

### Actual behavior:

https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/df0e154a-8fd4-4a3a-a239-d3ffe4cfa770

### Additional Details

Tested in different fresh sites with the next configs and in all of them it fails:

- WC 8.7-rc1 / WP 6.5-rc1
- WC 8.6.1 / WP 6.5-rc1
- WC 8.6.1 / WP 6.4

puntope:

Closes #377 and #381 

Handle the `add_to_cart` and `select_content` events for the block in the same way we do for classic pages. Add listeners to the component internals, as the Experimental Hooks stopped working.

1. Install Store Front
2. GO to shop page
   0. Verify `view_item_list` is tracked for all the products on the page
   1. `add_to_cart` is tracked when you click on "Add to cart" button
   4.  `select_content` & no `add_to_cart` is tracked, when
       1. Click on product image
       2. Click on the product name
       3. Click on "Select options"
       5. Click on "View products"

6. Create a Page and add Products Block (Beta) 
  Repeat and verify steps from step 2 

### Changelog entry

> Fix - Add to cart & select content events for All Products Block.


tomalec:
Thanks @puntope !
I tweaked a few more things:

- covered the cases when you click `event.target` is a descendant of a product button component.
- prevent double `add_to_cart` for classic listings
- add support for `select_content`  via clicking on the product image in the block listing
- separated the code, as IMHO it would make it easier to reason and more accessible to remove once Block events are back.

puntope:
HI @tomalec thanks for your contributions to the PR. I added some tweaks supporting clicking on the product name as well. 

This should be working now.

puntope:
```php
add_action(
   'woocommerce_shop_loop_item_title',
   function () {
      global $product;
      $this->append_script_data( 'products', $this->get_formatted_product( $product ) );
   }
);
```

This won't run for blocks. It will create a fatal in Product Blocks because products variable will be undefined when calling the tracker eventHandler 

I tweaked it with `woocommerce_loop_add_to_cart_link` here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/commits/805208c8867d9db1597cd1f89e08c72058b37c4e

puntope:
`view_item_list` was also not working. It was not added in the classic page either, so I wonder if we remove it intentionally?

IN any case I added support for it here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/commits/f29a44a8f50661a3f0089ac0a4cf1f99a7773eb8

mikkamp:
I think we need to take a few more things into consideration in this PR, as we have the following we need to support:

1. Traditional pages using the classic shortcodes.
2. Default WooCommerce pages which are created during install, most have been switched to blocks when starting with a new site.
3. The "Products (Beta)" block.
4. The "Product Collection (Beta)" block. This one is intended to replace the Products block.

One thing I noticed when testing those is the selector for `add_to_cart` in this PR needs to change to the following to support the Product Collection block:

```js
		.querySelectorAll(
			'.products-block-post-template .product, .wc-block-product-template .product'
		)
```

Also can you clarify exactly what is supposed to trigger the `select_content` event. I've tried testing it in various scenarios, but in most cases I'm just redirected to a new page before anything is tracked. Can you explain how I should be testing that?

assets/js/src/integrations/classic.js
```diff
@@ -156,4 +158,67 @@ export function trackClassicPages( {
...
+				}
+			} );
+		} );
+
+	// Track View Item List
+	if ( products?.length ) {
+		tracker.eventHandler( 'view_item_list' )( {
+			products,
+		} );
+	}
```
tomalec: :grey_question: 
The biggest concern I have is regarding this bit.
Other "on-page-load" events are handled through `events` object. I'm concerned not only about consistency, but also about firing an event twice.

I'd try reproducing and test how to achieve that live, but it would be nice to have E2Es in place https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/385

mikkamp: The E2E tests handle checking if the event is sent. But since it waits for all requests and then returns the one that matches. I haven't found a way yet to confirm it doesn't send an event twice.

puntope: I https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/files#diff-ca051807ffd7c39f888ccb191f645da9c2f795acc8f63eb4b0a87f6da5da0bbf `removeAction( `${ ACTION_PREFIX }-view_item_list`, NAMESPACE );` as well as https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/files#diff-feb1d32ea04490583d36ba93f2101db973aad5303bbabee4037af3767ac3711e `'view_item_list' => 'woocommerce_before_shop_loop_item',` from the mapping.

So I guess it won't trigger anywhere else that when product data is filled.

puntope:
> Also can you clarify exactly what is supposed to trigger the select_content event. I've tried testing it in various scenarios, but in most cases I'm just redirected to a new page before anything is tracked. Can you explain how I should be testing that?

Select Content will trigger when clicking when  interacting anywhere in the product except in the Add to cart button.

Examples:

- Clicking on the product picture 
- Clicking on the product name
- Clicking on "Customize" or any other button not "Add to cart" button 

After that event, normally the Shop goes to the Product Detail and the "view_item" event is triggered. It depends on the user journey but the flow until  the product is added to the cart might be:

- view_item_list
- select_content/add_to_cart
- view_item 
- add_to_cart

puntope:
> One thing I noticed when testing those is the selector for add_to_cart in this PR needs to change to the following to support the Product Collection block:

Thanks @mikkamp I added your suggestion here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/commits/193e06c2eace8c51b1ee7803f548c6206357bbb8

puntope:
1. Products Classic (adding [ products ] shortcode in a Page)

- Tested entering the page tracks "view_item_list" with all the items.
- Tested add to cart when adding a product to the cart
- Tested select_content when clicking on the product  picture
- Tested view_item triggers when accessing the product
- Tested the page tracks "view_item_list" with all the related items.
- Tested add to cart in the single product page

https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/c67df4ff-b1a1-445f-9499-5ad6e88d9c1c


2. Product  (Beta) Blocks (adding the "Product Beta Block" in a page)

- Tested entering the page tracks "view_item_list" with all the items.
- Tested add to cart when adding a product to the cart
- Tested select_content when clicking on the product  picture
- Tested view_item triggers when accessing the product
- Tested the page tracks "view_item_list" with all the related items.
- Tested add to cart in the single product page


https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/3593f7f1-db79-45fb-bbda-2b24155cbc1b


3. Products (default  TT4 theme products page)

- Tested entering the page tracks "view_item_list" with all the items.
- Tested add to cart when adding a product to the cart
- Tested select_content when clicking on the product  picture
- Tested view_item triggers when accessing the product
- Tested the page tracks "view_item_list" with all the related items.
- Tested add to cart in the single product page

https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/49163811-d7cf-4ff9-bf31-01d02435063b


Extra - Tested From Add to cart to Purchase


https://github.com/woocommerce/woocommerce-google-analytics-integration/assets/5908855/e5794830-b8a6-4984-aca8-b37dc7a9e829

puntope:
Hi @mikkamp and @tomalec 

I added the suggestion mentioned here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1989256479 and answered some of the pending topics here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1989254226 and here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#discussion_r1519464543.

CHeck also a full test recording here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1989290848

Product Blocks
Products (default  TT4 theme setup)
Products Classic 

I tested it in WP6.5-RC1 + WC 8.7.0-RC1

Can you please verify? Maybe with WP6.4 and WC8.6.2

Thanks

assets/js/src/integrations/classic.js
tomalec: I mean that we already have a logic to track events on page load. It's 
https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/c34d9325a9eca2aa7b93b78c3f8c1bd9d6e29594/assets/js/src/integrations/classic.js#L32:L45

Adding those few lines adds yet another place with almost the same responsibility, which would lead to future duplications, as we may miss one or the other.

I think we could avoid adding this bit in JS, and simply bring back `woocommerce_loop_add_to_cart_link` to the mappings list, or add it manually on the `woocommerce_loop_add_to_cart_link` callback. If we're adding a product to `products`, we could as well at the same time add `view_item_list` to `events`.

This way we will have one predicate to determine what to call on the pageload: "an event from `events` array".

tomalec: Like in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/commits/6ad4979e152f9986e551889042e27bd7d435ef14

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -87,11 +87,13 @@ function ( $cart_item_key, $product_id, $quantity, $variation_id, $variation ) {
...
-			function () {
-				global $product;
+			'woocommerce_loop_add_to_cart_link',
+			function ( $button, $product ) {
 				$this->append_script_data( 'products', $this->get_formatted_product( $product ) );
-			}
+				return $button;
+			},
+			10,
+			3
```
mikkamp: This should be a 2, since we only have 2 arguments we accept.

includes/class-wc-google-gtag-js.php
```diff
@@ -26,7 +26,7 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 	private $mappings = array(
 		'begin_checkout'   => 'woocommerce_before_checkout_form',
 		'purchase'         => 'woocommerce_thankyou',
-		'view_item_list'   => 'woocommerce_before_shop_loop_item',
+		'view_item_list'   => 'woocommerce_loop_add_to_cart_link',
```
mikkamp: Changing this breaks the shop pages, because we switch it from an action to a filter. While actions and filters are interchangeable in most cases the important difference is that a filter must return a value otherwise the next filter will get a blank value. This causes the add to cart buttons to disappear.

We need to either pick an action here. Or handle this separately so we can use `add_filter` in the function `map_actions`.

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -87,11 +87,13 @@ function ( $cart_item_key, $product_id, $quantity, $variation_id, $variation ) {
 		);
 
 		add_action(
-			'woocommerce_shop_loop_item_title',
-			function () {
-				global $product;
+			'woocommerce_loop_add_to_cart_link',
```
mikkamp, line 90: We are switching from an action to a filter here. So we should use `add_filter` to explain why we need to return the button value.

mikkamp (commented):
The last commit breaks my shop pages, by removing the add to cart buttons. We might need to expand mapping to support both actions and filters.

mikkamp:
Unit tests are failing because WP downloads from https://wordpress.org are temporarily unavailable. Can retry after that's fixed, but I've tested it locally and it passes.

Update: Downloads were fixed, so re-ran the tests. They all passed now.

puntope:
Hi @mikkamp Thanks for the last tweaks. 

I tested everything again with WC 8.6 as well as WC 8.7 and followed the steps here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1989290848 and is working as expected.

I guess this is ready for approval

puntope:
Added support for All Products Block. (Needs to create a page and add All Products Block on it)

That requires to rollback the deletion of the unique actions.  

See https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/commits/ee2740db6c53987c4aeffd831a08da8770f15911 and https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378/commits/fb2b86240019e42ed93f9d9f33c792a7bb50559c

mikkamp (approved):
LGTM, thanks for all the revisions.

Issue #368, opened 2024-03-06
Latest Update Issue


kingyyyyyy:
### Describe the bug:
Hello,

We see the following error in console, with the new update:-

https://i.gyazo.com/8ae21a0b4145bbf6d5fadfc38e30b55e.png

https://i.gyazo.com/466b3ce707143009319b601167f7fa56.png

We have shown our settings below.

https://i.gyazo.com/afa55784663023309f4a303e01da4016.png

Also, we no longer see most visitors on Real Time on GA (started happening around 2 hours ago)? We have orders coming through and have seen only 2 real time visitors in that time, even though earlier on it was over 20 constantly. 

Tested using Tag Assistant Companion extension and it can’t find a tag on our site, too, anymore?

We don't use Google Ads etc, due to the nature of our business, and I believe this is the only plugin we use for GA.

All plugins on the site are updated.

Any idea how to fix this?

### Expected behavior:
Works like before

### Actual behavior:
Issues outlined above since plugin update and cache clearing.

### Additional details:
`
### WordPress Environment ###

WordPress address (URL): https://myeliquidsupplies.com
Site address (URL): https://myeliquidsupplies.com
WC Version: 8.6.1
REST API Version: ✔ 8.6.1
WC Blocks Version: ✔ 11.8.0-dev
Action Scheduler Version: ✔ 3.7.1
Log Directory Writable: ✔
WP Version: 6.4.3
WP Multisite: –
WP Memory Limit: 512 MB
WP Debug Mode: –
WP Cron: ✔
Language: en_GB
External object cache: –

### Server Environment ###

Server Info: Apache/2.4.58
PHP Version: 7.4.33
PHP Post Max Size: 120 MB
PHP Time Limit: 300
PHP Max Input Vars: 10000
cURL Version: 7.68.0
OpenSSL/1.1.1f

SUHOSIN Installed: –
MySQL Version: 8.0.36-0ubuntu0.20.04.1
Max Upload Size: 120 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

WC Database Version: 8.6.1
WC Database Prefix: wp_
Total Database Size: 1278.56MB
Database Data Size: 813.75MB
Database Index Size: 464.81MB
[...]

PR #379, opened 2024-03-08
Warn about potential scripts order mismatch.

tomalec:

Warn about potential scripts order mismatch.
to make it easier to track issues like https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/368

### Checks:

1. Build, install activate.
2. Install and activate Autoptimize plugin
3. Set it up
   - _Do not aggregate but defer?_ - enabled
   - _Also defer inline JS?_ disabled
   - _Exclude scripts from Autoptimize:_ empty
   image

### Changelog entry

> Tweak - Warn about potential script order mismatch

### DIFF
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -105,11 +105,23 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 			)
 		);
 
+		$message = sprintf(
+			/* translators: 1 inlined data script, 2 the tracker definition */
+			__( "`wcgai.trackClassicPages` function is not available. Please make sure Google Analytics for WooCommerce's scripts are loaded correctly. `%1\$s` after `%2\$s`. Some other plugins may defer or change the script order.", 'woocommerce-google-analytics-integration' ),
+			$this->data_script_handle,
+			$this->script_handle,
+		);
+
 		wp_add_inline_script(
 			$this->data_script_handle,
 			sprintf(
-				'wcgai.trackClassicPages( %s );',
-				$this->get_script_data()
+				'if( window.wcgai && typeof window.wcgai.trackClassicPages === "function" ){
+					wcgai.trackClassicPages( %s );
+				} else {
+					console.error( "%s" );
+				}',
+				$this->get_script_data(),
+				esc_js( $message ),
 			)
 		);
 


includes/class-wc-google-gtag-js.php
```diff
@@ -105,11 +105,23 @@ public function inline_script_data(): void {
...
 		wp_add_inline_script(
 			$this->data_script_handle,
 			sprintf(
-				'wcgai.trackClassicPages( %s );',
-				$this->get_script_data()
+				'if( window.wcgai && typeof window.wcgai.trackClassicPages === "function" ){
+					wcgai.trackClassicPages( %s );
+				} else {
+					console.error( "%s" );
+				}',
```
martynmjones, line 122: It looks like there are more issues with optimisation plugins than originally thought so I was wondering about adding a fallback. Instead of showing the console error if `wcgai.trackClassicPages` is unavailable, what if we then add a `DOMContentLoaded` event listener and try once more? Something along the lines of

```js
				'(function(){
					const config = %s;
					const didInit = () => {
						return window.wcgai && typeof window.wcgai.trackClassicPages === "function";
					}

					if( didInit() ){
						wcgai.trackClassicPages( config );
					} else {
						document.addEventListener( "DOMContentLoaded", () => {
							if ( didInit() ) {
								wcgai.trackClassicPages( config );
							} else {
								console.error( "%s" );
							}
						});
					}
				})();',
```

I tested that with Autoptimize which is one that breaks completely and that resolved the problem.

We didn't originally use the `DOMContentLoaded` event because it's pushing the initialization later and that could cause issues with other plugins or customisations but it does mean the core tracking would work as expected. We could also add a notice at the second attempt to `wcgai.trackClassicPages()` that states third-party functionality may not work as expected.

What do you think?

tomalec: Thanks for catchign that.

I have a bit mixed feelings. one one hand it's nice to give a fallback and the least degradaded experience possible. On the other, people may just live with that alternative behavior, relying on that. Which will not work well with the consent mode (it should be set before any events), plus they would expect us to support it.

martynmjones: Yeah I understand what you mean, however, with the Consent mode changes in #364 merchants can set the defaults using the filter and it's included in the config. So even running `trackClassicPages` on `DOMContentLoaded` the default consents are all sent before other events.

It's not a great solution but IMO the benefit would outweigh the cons with how many merchants would fall within each of those possibilities.

tomalec: Meaning: to guarantee plugin's functionality and support for the consent mode, we need to make sure tracker is initialized synchronously in a blocking script, before we fire any events.

Also, do I understand correctly that people are reporting issues that JS behavior is broken with the plugins that explicitly break the WorPResses dependency management? 
I think we should try our best to provide the bullet-proof functionality, but I think we cannot anticipate and cover cases where third parties deliberately break WP features.

I have two alternative ideas:

1. We could just step out of the current structure of scripts, and  maybe use rely on in-browser dependency management with ES Modules. That could solve the `trackClassicPages` + `config` issue, but we still need to make sure the tracker blocking stript was loaded first. So we could add the warning to the `tracker.js` to check if it was executed before `documentloaded`. Or simply inline the tracker's definition.
2. We could avoid/move the `domcontentloaded` logic to the `trackClassicPages`, so we would be sure the function is there 
   - inline script:
      - Adds event data to a global `wcgai` object or to `currentScript` (having data in a global could make debugging easier)
      - check if there's `trackClassicPages`, calls it, otherwise triggers a warning
   - `classic.js`
      - defines the `window.trackClassicPages`
      - check if there's `window.wcgai.data` defined, call `trackClassicPages(data)` and trigger warning, otherwise do nothing.
      
   This way we don't need to wait for domconentloaded, we simply rely on the dualism, either
    - `woocommerce-google-analytics-integration` is loaded before `woocommerce-google-analytics-integration-data` as it should, or
    - it is loaded after, then we have data and the function in place to fallback and throw a warning
    - if any of two is missing there is nothing we could do.

mikkamp: Correct me if I have any of the logic wrong but can't we KISS it by splitting the tracker code to a separate script (at least the constructor part, event tracking can still stay in main.js). Then we can do everything possible to ensure that part loads early. Autoptimize has the `<!--noptimize-->` tag and other plugins have an attribute you can set to prevent it from being optimized.

As for the rest of the main.js script and the data it requires we can keep that simple by having fallbacks to support any loading order, for example:
```
<script for data>
  set data in window.wcgai.data
  if window.trackClassicPages exists call it with set data
</script>
<script for main.js>
  construct and initialize window.trackClassicPages
  if window.wcgai.data exists and trackClassicPages is not called yet, then call it here
</script>
```
These last two scripts can then be optimized and deferred as they like without any consequences.

We would be breaking the comment you suggested here:
> tracker.js needs to be executed ASAP, the remaining bits for main.js could be deffered, but to reduce the traffic, we ship it all together

However that tracker code is going to be very small and it won't change so browsers can cache it after the first load.

I think that might leave is with one additional problem if the tracker script and it's data swap order, but we could do the same as with main.js to make the load order be swappable. And if we don't want to use too many globals we can place everything under `window.wcgai` to have less chance of conflicts.

martynmjones: > Correct me if I have any of the logic wrong but can't we KISS it by splitting the tracker code to a separate script (at least the constructor part, event tracking can still stay in main.js). Then we can do everything possible to ensure that part loads early. Autoptimize has the  tag and other plugins have an attribute you can set to prevent it from being optimized.

I agree that we should separate the tracking initialisation and event tracking. Combining the two was a misstep and has caused more problems than is solves and means that the extension is far less resilient against optimisation plugins that might not fully adhere to dependencies and loading order.

Me and Tomek were discussing a few different ideas last night and after giving it more thought I think we should take a step back to cover more scenarios:

1. Remove the tracking function setup from our JS files
2. Hook into `wp_head` early and output the Google Tag Manager script along with tracker setup directly
    - While it's not the WordPress way because we wouldn't be using `wp_enqueue_script`, it would be far more reliable and we wouldn't then have any issues with optimisation plugins moving things around
    - We would still need to register the `google-tag-manager` handle to prevent Core from enqueuing it's own script (_I'm planning on opening a PR to remove the integration in core_)
    - Added benefit that we would be following Google's installation direction exactly regardless of site setup "_Place the Google tag snippet immediately after the opening `<head>` HTML tag on every page you want to measure._"
3. Enqueue `main.js` but don't run any code for the integration until `DOMContentLoaded`
4. Add `wcgaiData` via inline script

If we were to do that then we couldn't need to worry about the tracker being available at any point or to other extensions, Consent defaults would always be sent before any other calls, and we wouldn't need to do anything about any optimisation plugins because the loading order of our event tracking would become irrelevant.

martynmjones: I've opened a PR which is loosely based on something I previously did with the added separation of `gtag` setup and event tracking as a demonstration #398 

This resolves any problems with 3PD plugins changing file load order and prevents anything from executing before the config is accessible. Interested in hearing thoughts or if any one has alternative suggestions that would be more suited?

tomalec:
Most probably we will close this PR in favor of making the code more resilient for scripts reordering in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398

PR #380, opened 2024-03-08
Format price for purchase event

tomalec:

Format prices for `purchase` events the same way we do for other events.
Fix https://wordpress.org/support/topic/conversion-value-wrong/

### Checks:

1.  Build, install & activate
2. Use [Tag Assistant](https://tagassistant.google.com/)
3. (setup your store to use decimal part in prices)
4. Open shop page, add item to the cart
5. Place an order
6. Inspect `purchase` event in Tag Assistant
7. Check that prices of individual items and the entire cart ar in the same format

#### Before
image
#### After 
image

### Changelog entry

> Fix - Purchase price format.

### DIFF
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -178,8 +178,11 @@ export const purchase = ( { order } ) => {
 	}
 
 	return {
-		currency: order.currency,
-		value: parseInt( order.value ),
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
Thanks for the changes, this is working perfectly for me now.

I was experimenting with some local E2E testing and the test that was previously failing now passes after this PR:
image

PR #384, opened 2024-03-11
Get correct variation data when formatting product

mikkamp:

The function `get_formatted_product` is used to get consistent product data and is called in various ways. This highlights two problems:

1. When we call it from `woocommerce_add_cart_item` we pass the variation data but the product we pass is the parent product, which means we don't get the price of the variation but instead from the variable product (which results in the lowest variation price being tracked)
2. When we call it with a product that is a variation, we fetch the parent ID to track. But the categories are still fetched from the product object (variations can't have categories assigned), we should be getting categories from the parent product.

This PR addresses those two issues.

#### Scenario for variation price
1. Set up the basic tracking (including add_to_cart tracking)
2. Create a variable product (with different prices per variation) 
3. Add the variable product to the cart choosing a variation that is not the lowest price (in an incognito window to make sure it's tracked)
4. Check the tracked data (either through a browser plugin or view network requests and filter for `collect?`
5. Ensure tracking data has the right price, example:
```
en=add_to_cart&_ee=1&pr1=id85~nmVariable~caUncategorized~qt1~pr6~vapa_color%3A%20red

pr1 = data for product 1 (separated by `~`, first 2 characters are key)
pr6 = price 6.00
``` 

#### Scenario for variation categories
1. Set up the basic tracking (including add_to_cart tracking)
2. Create a variable product
3. Add the variable product to the cart (in an incognito window to make sure it's tracked)
4. Go to the checkout page and check for the `begin_checkout` event
5. Check the tracked data (either through a browser plugin or view network requests and filter for `collect?`
6. Ensure tracking data has the right categories (Uncategorized if no categories were assigned), example:
```
en=begin_checkout&_ee=1&pr1=id85~nmVariable~caUncategorized~qt3~pr12~vapa_color%3A%20green

pr1 = data for product 1 (separated by `~`, first 2 characters are key)
caUncategorized = category Uncategorized
``` 

### Additional details:
One followup thought this leaves me with, in the case where we track through `woocommerce_add_cart_item`, should we be getting the price from the cart data or from the product? If we've applied a discount then the cart data would seem more accurate since it would be the final price they are expecting to pay.

We can also notice that the `begin_checkout` logs the price x quantity, whereas the purchase event logs the individual product price and the quantity separately (we should be consistent there).

### Changelog entry
* Fix - Get correct variation data when formatting product.


tomalec (approved):
Reviewed the code, tested locally LGTM

PR #385, opened 2024-03-11
Add E2E testing to track events across all page types

mikkamp:

With the focus on shifting towards blocks yet still supporting classic pages, we have to cover the following types of pages to track their events:
1. Traditional pages using the classic shortcodes.
2. Default WooCommerce pages which are created during install, most have been switched to blocks when starting with a new site.
3. The "Products (Beta)" block.
4. The "Product Collection (Beta)" block. This one is intended to replace the Products block.

To test this manually is a large burden. This PR copies over the E2E we have for tracking in GLA and modifies them to work with the GA events we send. This makes it easier to test the events across several different page/block types.

The tests in this PR did highlight some inconsitencies, and events that aren't being tracked, so the following changes are needed to make these tests pass:
- PR #384 
- PR #378 
- Extra selector for Product Collection block mentioned in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1987786319

1. Install any required patches as listed above
2. Install all npm dependencies  `nvm use && npm ci`
3. Build assets either through `npm start` or `npm run build`
4. Ensure we don't have any conflicting E2E docker instances running (like the one from GLA)
5. Start the wp-env site `npm run wp-env:up`
6. Install browser (only needed once) `npx playwright install chromium`
7. Run tests `npm run test:e2e`
8. Confirm tests are successful

Note: workflow is included (direct copy from GLA). We can only trigger it once this PR is merged (unless you want to clone the repo to test it).

### Additional details:

These test aim to cover the most common scenarios, but while testing I still came across the following inconsistencies/parts we aren't covering (these can be addressed in follow up PR's):

- [ ] Remove from cart with shortcode `[woocommerce_cart]` isn't always tracked because of immediate page redirect (not in E2E)
- [ ] Remove from cart / Begin checkout is not consistent between blocks and shortcode, category and variation data differ.
- [ ] Change cart quantity is not tracked with shortcode `[woocommerce_cart]` or with blocks (PR #375 removes it)
- [ ] Add shipping info doesn't seem to be used anywhere (mentioned in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/375#issuecomment-1987887419)
- [ ] Select content is difficult to test if it is followed up by a page redirect (no E2E tests written yet)
- [ ] `begin_checkout` logs total item price (including discounts and price * quantity), `purchase` logs product object price (not final price we paid in the order). Should these be consistent?
- [ ] When we hook into `woocommerce_add_to_cart`, should we get final price from cart_item_data? Otherwise we'll track product prices vs. discounted price.
- [ ] When we change the settings to redirect to cart page after `add_to_cart`, do we still capture all details?
- [ ] Testing with grouped / bundled products (grouped product seems to only track first product added to cart)

### Changelog entry
- Dev - Add E2E testing to track events across all page types.


tests/e2e/utils/create-page.js
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Creates a classic cart page using shortcodes.
+ */
+export async function createClassicCartPage() {
+	const title = 'Classic Cart';
+	const content = '[woocommerce_cart]';
+
+	if ( ! ( await pageExistsByTitle( title ) ) ) {
+		await createPage( title, content );
+	}
```
tomalec: :scroll:  :nail_care: 
If we're going to extract those to shared lib, we could consider all `create*Page` functions to return the created page id. to keep their behavior consistent

ianlin (commented):
I only verified if all the e2e tests were passed. Here's my step:

- Run e2e tests on this branch, most of the tests failed which is expected
- Check out a new branch based off this PR's branch `dev/add-e2e-tests`
- Merge `origin/fix/add_to_cart_selector` from #378 into the new branch
- Merge `origin/fix/get-data-for-variations` from #384 into the new branch
- Update code locally based on https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1987786319
    ```diff
    --- a/assets/js/src/integrations/classic.js
    +++ b/assets/js/src/integrations/classic.js
    @@ -162,7 +162,9 @@ export function trackClassicPages( {
      // Handle selection and add_to_cart in **Block** product listing the classic way.
      // Attach click event listeners to a whole product card, as some links may not have the product_id data attribute.
      document
    -   .querySelectorAll( '.products-block-post-template .product' )
    +   .querySelectorAll(
    +     '.products-block-post-template .product, .wc-block-product-template .product'
    +   )
        ?.forEach( ( productCard ) => {
          // Get the Product ID from a child node containing the relevant attribute
          const productId = productCard
    ```
- Run e2e tests again, all tests passed except:
    ```
    6 [chromium] › gtag-events/blocks-pages.test.js:160:2 › GTag events on block pages › Add to cart event is sent from a product collection block shop page
    ```
    Which is related to the code change from https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1987786319. Not sure if I change the code locally wrong?

I saw @tomalec started reviewing this PR as well I would leave the rest of the review to him for now. I would review the rest tomorrow if needed.

tests/e2e/utils/customer.js
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Add a product to the cart from a block shop page.
+ *
+ * @param {Page}   page
+ * @param {number} productID
+ */
+export async function blockProductAddToCart( page, productID ) {
+	const addToCart = `[data-product_id="${ productID }"]`;
```
tomalec, line 88: :nail_care: :nail_care: 
Depending on the product, this selector could match not the "Add to cart" button but the "Select option" button.

From the implementation behavior perspective, I think it's fine. The user of the function probably is aware of what product ID they give, but maybe it could be worth mentioning in a comment.

mikkamp:
> Run e2e tests again, all tests passed except:
```
[chromium] › gtag-events/blocks-pages.test.js:160:2 › GTag events on block pages › Add to cart event is sent from a product collection block shop page
```
> Which is related to the code change from https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378#issuecomment-1987786319. Not sure if I change the code locally wrong?

The code you changed is right, did you rebuild the assets after that change? I'm not quite sure why it would be failing otherwise. Maybe you can check the site at `http://localhost:8889` to see how it handles the "Product Collection Shop" page. Make sure to add some test settings to enable tracking, during E2E that's enabled and cleared.

tests/e2e/specs/gtag-events/blocks-pages.test.js
```diff
@@ -0,0 +1,271 @@
...
+
+		// Go to shop page (newest first)
+		await page.goto( 'shop?orderby=date' );
+		await blockProductAddToCart( page, simpleProductID );
+
+		await event.then( ( request ) => {
+			const data = getEventData( request, 'add_to_cart' );
+			expect( data.product1 ).toEqual( {
+				id: simpleProductID.toString(),
+				nm: 'Simple product',
```
tomalec, line 54: :nail_care: Maybe instead of hardcoding its name, we could read it from `config.products.simple.name`?

tests/e2e/utils/customer.js
mikkamp: Definitely agree, but couldn't find another selector that's identical for both the `Products` block and the `Product Collection` block. I was considering using `product_type_simple` but then we'd limit the helper function to one specific product type.

It's pretty much a copy of what we have in GLA and since we always call it with a specifically prepared productID we know what to expect when using the function. So I'm just going with adding the comment.

tests/e2e/specs/gtag-events/blocks-pages.test.js
```diff
@@ -0,0 +1,271 @@
...
+
+const config = require( '../../config/default' );
+const simpleProductPrice = parseFloat( config.products.simple.regular_price );
+
+let simpleProductID;
+
+test.describe( 'GTag events on block pages', () => {
+	test.beforeAll( async () => {
+		await setSettings();
+		simpleProductID = await createSimpleProduct();
```
tomalec: :nail_care: 
As we use `simpleProductID` only in the scope of `GTag events on block pages` and do not clear it after, I'd move the declaration there. To avoid artifacts when we add another `describe`.
```suggestion
test.describe( 'GTag events on block pages', () => {
	let simpleProductID;
	
	test.beforeAll( async () => {
		await setSettings();
		simpleProductID = await createSimpleProduct();
```

tests/e2e/specs/gtag-events/blocks-pages.test.js
tomalec: :nail_care: Also, we could cache `simpleProductID.toString()` and `simpleProductPrice.toString()`in beforeAll

mikkamp: I agree, but also disagree here. I personally don't like to abstract too much information in the `expect` calls. I want to see clearly what data I'm expecting without having to look it up what it should match too. It's true that if they are constants then there is no harm as they can't be modified, but this is more personal preference. Ofcourse I'm breaking my own by not using a constant for simpleProductPrice, but that's mainly because I copied that logic from GLA.

For the same reason I don't often make a lot of the expect parts in tests "DRY", which could easily be done in this case since a lot of expect similar data.

tomalec: I can relate to being explicit in `expect`s, so I'm fine leaving it as `simpleProduct*.toString()`.

But for the same reason, I'd use `config.products.simple.name` as this is what we expect from the perspective of this file. We didn't set `Simple product` here. So we expect it to be "whatever is set in config"

tests/e2e/utils/create-page.js
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * @return {number} Created page ID.
+ */
+export async function createClassicCartPage() {
+	const title = 'Classic Cart';
+	const content = '[woocommerce_cart]';
+
+	if ( ! ( await pageExistsByTitle( title ) ) ) {
+		return await createPage( title, content );
+	}
```
tomalec: We're not returning the ID if the page exists.
```suggestion
	return (
		( await pageExistsByTitle( title ) ) ||
		( await createPage( title, content ) )
	);
```

tomalec (approved):
Reviewed the code, run locally, LGTM
Left a few cosmetic comments.

Thank you very much! :) Having E2Es in place could boost the speed and confidence in our development :)

tests/e2e/utils/__fixtures__/product-collection.fixture.json
```diff
@@ -0,0 +1 @@
...Loop\":true,\"textAlign\":\"center\",\"fontSize\":\"small\"} /-->\n\n<!-- wp:woocommerce/product-button {\"textAlign\":\"center\",\"isDescendentOfQueryLoop\":true,\"fontSize\":\"small\"} /-->\n<!-- /wp:woocommerce/product-template -->\n\n<!-- wp:query-pagination {\"layout\":{\"type\":\"flex\",\"justifyContent\":\"center\"}} -->\n<!-- wp:query-pagination-previous /-->\n\n<!-- wp:query-pagination-numbers /-->\n\n<!-- wp:query-pagination-next /-->\n<!-- /wp:query-pagination -->\n\n<!-- wp:woocommerce/product-collection-no-results -->\n<!-- wp:group {\"layout\":{\"type\":\"flex\",\"orientation\":\"vertical\",\"justifyContent\":\"center\",\"flexWrap\":\"wrap\"}} -->\n<div class=\"wp-block-group\"><!-- wp:paragraph {\"fontSize\":\"medium\"} -->\n<p class=\"has-medium-font-size\"><strong>No results found</strong></p>\n<!-- /wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p>You can try <a href=\"#\" class=\"wc-link-clear-any-filters\">clearing any filters</a> or head to our <a href=\"#\" class=\"wc-link-stores-home\">store's home</a></p>\n<!-- /wp:paragraph --></div>\n<!-- /wp:group -->\n<!-- /wp:woocommerce/product-collection-no-results --></div>\n<!-- /wp:woocommerce/product-collection -->"}
```
tomalec: :grey_question:  :nail_care: 

Why do we use `/__fixtures__/` insted of `/fixtures/`? Do we autogenerate them somehow? If not, maybe we could format  the content to be more readable.

mikkamp: No specific reasoning, I manually generated these.
Copied directly from GLA, where we initially copied it from the WC Blocks E2E tests. So I'm not tied to a particular method, I just needed something up and running.

mikkamp:
@tomalec Thanks for the review. Since this is approved and working, I'm going to merge it so we can focus on the other issues and get them tested. We can always refactor / change any other details later if needed.

PR #386, opened 2024-03-11
Handle AJAX add to cart for simple products

### LINKED ISSUE #382: CRITICAL BUG in new tracking implementation (v2.0.1)
### Describe the bug:

When a Product is added or removed from Cart, the tracking script relies on the add_to/remove_from cart button of that product_id to be part of the current DOM, otherwise a JS error is triggered.

COMMON WAYS THIS ASSUMTION IS WRONG:
Grouped, Bundled, Chained Products, BOGO Deals, Free Products you name it..... which Plugin is used or if it is done by code doesn't matter, whenever a product where the add_to/remove_from cart button is not part of the DOM, is added/removed from cart we get a major crash.

This is extremely critical as it breaks compatibility with 100s of extensions, even the ones from Automattic/Woocommerce itself.

There is no way to avoid the added_to_cart/remove_from_cart events to trigger other than editing your js files. As the event handlers are hardcoded in js and recieve events EVEN IF tracking added_to_cart/remove_from_cart events is disabled via the settings screen.

### Steps to reproduce:
1. Setup Woocommerce
2. Install ANY Extension from Automattic/Woocommerce or a 3rd Party vendor that adds Products to cart or removes them, that don't rely on an add_to_cart or remove_from_cart button.

If you want to test it via a Plugin, here are some Examples:
Woocommerces build-in Grouped Products
Product Bundles by WooCommerce
Composite Products by WooCommerce
Any BOGO Deals / Coupons Plugin by Woocommerce

4. Trigger add_to_cart/remove_from_cart on that product not part of the current html markup.

ALTERNATIVE

1. Setup Woocommerce
2. trigger add_to_cart or remove_from_cart on a product that is not part of the current html markup.

### Expected behavior:
1. The add_to_cart/remove from cart event is triggered as usual with the product_id. Or if the product id is not found on that page the event is simply ignored.

2. if added_to_cart/remove_from_cart events is disabled via the settings screen the classic.js implementation should not listen to them.

### Actual behavior:
JS error and breakage of the page.
The crash happens in any of the functions of assets/js/src/utils/index.js, cause the passed product is null/undefined.
Why is it null, cause it is not part of the products array in assets/js/src/integrations/classic.js, that is used to fetch the product.

BUT: The actual issue is that the plugin hooks into add_to_cart/remove_from_cart actions and triggering all the js events and relying on the product add/remove button being part of the DOM, which is not the case in a number of cases as explained before.

### Additional details:
We really need to get this fixed cause right now, nearly every woocommerce site out there is broken caused by the change in tracking.

In short: what you indent to track via js is add_to/remove_from cart button clicks, but what you are actually tracking is add_to/remove_from cart events - via woocommerce hooks - which is a total different thing :-/ so you cannot rely on a button being in DOM.

I provided a quick solution in the comment below! would be nice if someone @tomalec @puntope @martynmjones can have a look at this!

mikkamp:

This PR resolves the issue that an add to cart AJAX request from a single product page was throwing an error.
WC core doesn't handle AJAX from single product pages, however it can be replicated with [a plugin](https://wordpress.org/plugins/woo-ajax-add-to-cart/). It's also quite common for themes to include this kind of behaviour.

Reported in the forum: https://wordpress.org/support/topic/js-error-on-product-page-2/

The PR resolved two issues:
1. If the product data attribute is not available on the add to cart button, then extract it from the value (WC sets it on the value in [this template](https://github.com/woocommerce/woocommerce/blob/8.6.1/plugins/woocommerce/templates/single-product/add-to-cart/simple.php#L49)).
2. The handlers / formatter is given the result of searching for a product. However in some cases a product is not found by the function `getProductFromID`, in such cases we need to prevent from formatting the data.

We also need to be careful with the handler `document.body.onadded_to_cart` as we always hook into this regardless if we disable `add_to_cart` tracking, should we conditionally hook into that event?

Closes #382 

1. Setup tracking including `add_to_cart` tracking
2. Install a plugin to do AJAX add to cart: https://wordpress.org/plugins/woo-ajax-add-to-cart/
3. Create a simple product
4. Add simple product on a single product page
5. Confirm that the product is tracked
6. Create a page with the shortcode `[products]` and ensure AJAX add to cart on archives is enabled in WC > Settings
7. Add a simple product from the archive page
8. Confirm that the product is tracked

### Additional details:

This PR points out a separate issue that we search for products in the cart items first and then the list of products, [see here](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2.0.1/assets/js/src/utils/index.js#L167-L168).

That's a problem when it comes to the quantity number, which we can see in the following scenario:
1. Add single product 3x to cart
2. Refresh archive page
3. Add same single product to cart
4. View tracking data and the quantity is set to 3

Ideally we should be setting the quantity back to 1. Also is it really relevant to search through cart data, even though the product ID matches, we might have added one with a discount / addon which could make the price different. We'll need to log this in a separate issue.

Another issue is that it still won't work for variable products added to AJAX as we'll need to access the form data to find out which variation was selected.

### Changelog entry
* Fix - Handle AJAX add to cart for simple products

### DIFF
--- a/assets/js/src/integrations/classic.js
+++ b/assets/js/src/integrations/classic.js
@@ -52,13 +52,23 @@ export function trackClassicPages( {
 	 * @param {HTMLElement[]} button    - An array of HTML elements representing the add to cart button.
 	 */
 	document.body.onadded_to_cart = ( e, fragments, cartHash, button ) => {
-		tracker.eventHandler( 'add_to_cart' )( {
-			product: getProductFromID(
-				parseInt( button[ 0 ].dataset.product_id ),
-				products,
-				cart
-			),
-		} );
+		// Get product ID from data attribute (archive pages) or value (single product pages).
+		const productID = parseInt(
+			button[ 0 ].dataset.product_id || button[ 0 ].value
+		);
+
+		// If the current product doesn't match search by ID.
+		const productToHandle =
+			product.id === productID
+				? product
+				: getProductFromID( parseInt( productID ), products, cart );
+
+		// Confirm we found a product to handle.
+		if ( ! productToHandle ) {
+			return;
+		}
+
+		tracker.eventHandler( 'add_to_cart' )( { product: productToHandle } );
 	};
 
 	/**
--- a/assets/js/src/integrations/classic.js
+++ b/assets/js/src/integrations/classic.js
@@ -59,7 +59,7 @@ export function trackClassicPages( {
 
 		// If the current product doesn't match search by ID.
 		const productToHandle =
-			product.id === productID
+			product?.id === productID
 				? product
 				: getProductFromID( parseInt( productID ), products, cart );
 
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -46,7 +46,7 @@ export const view_item_list = ( {
  */
 export const add_to_cart = ( { product, quantity = 1 } ) => {
 	return {
-		items: [ getProductFieldObject( product, quantity ) ],
+		items: product ? [ getProductFieldObject( product, quantity ) ] : [],
 	};
 };
 
@@ -59,7 +59,7 @@ export const add_to_cart = ( { product, quantity = 1 } ) => {
  */
 export const remove_from_cart = ( { product, quantity = 1 } ) => {
 	return {
-		items: [ getProductFieldObject( product, quantity ) ],
+		items: product ? [ getProductFieldObject( product, quantity ) ] : [],
 	};
 };
 
@@ -128,6 +128,10 @@ export const add_shipping_info = ( { storeCart } ) => {
  * @param {Object} params.product The product to track
  */
 export const select_content = ( { product } ) => {
+	if ( ! product ) {
+		return false;
+	}
+
 	return {
 		content_type: 'product',
 		content_id: getProductId( product ),


puntope:
Thanks @mikkamp  ✅  LGTM

> Create a page with the shortcode [products] and ensure AJAX add to cart on archives is enabled in WC > Settings
> Add a simple product from the archive page
> Confirm that the product is tracked

These 3 steps are also fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378 (or at least I cannot reproduce them when using that branch) So I tested everything with that branch merged as well to be sure is not a conflict and is handled twice.

puntope (approved):

mikkamp:
> These 3 steps are also fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/378 (or at least I cannot reproduce them when using that branch) So I tested everything with that branch merged as well to be sure is not a conflict and is handled twice.

Thanks for double checking. The adjusted logic makes it hard to reproduce after #378, but the important part of this PR is that it also falls back to getting the product ID from the value of a button (which is where WC places it for simple products).

Since this is approved I'm going to go ahead and merge it.

lith-imad:
> Another issue is that it still won't work for variable products added to AJAX as we'll need to access the form data to find out which variation was selected.

I think this is a critical bug and should be fixed in its own issue. For anyone who is looking for a quick fix, here's what I've done as workaround: 

```
var addToCartButton = document.querySelector('.single_add_to_cart_button');
if (addToCartButton) {
	addToCartButton.dataset.product_id =  addToCartButton.parentElement.querySelector('input[name="product_id"]')?.value;
}
```

PR #388, opened 2024-03-11
Track purchase transaction ID

### LINKED ISSUE #387: Upgrading to 2.0.0 - Purchase Events Not Including Transaction ID
### Describe the bug:
Since upgrading to 2.0.0 last week we no longer see specific Transaction IDs being sent to GA4, and conversion data is significantly overstated. All purchase events are being sent with a value and items, but the Order ID is not being sent.

### Steps to reproduce:
1. Complete a purchase on a site using GA4 transaction tracking.
2. Observe the collect event sent to GA4.

### Expected behavior:
The purchase event sent to GA4 should include a unique transaction identifier, previously Order ID was sent under the UA configuration.

### Actual behavior:
Transaction ID is no longer sent, only currency, items, and value.

### Additional details:
~~Conversion amounts seem to be drastically overstated since the change as well, although this could be due to the nature of not having a unique transaction identifier within GA4 so each purchase is additive.~~
I found the note on a forum post to revert to 1.8.14 due to incorrect values being posted. However, I believe this missing Transaction ID reference should be included in the fix for that (#380).

mikkamp:

This PR adds the transaction ID (order ID) which is missing from the purchase event. Previously we sent the following details that [can be seen here](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/1.8.14/includes/class-wc-google-gtag-js.php#L342-L353):
```php
'transaction_id' => $order->get_order_number(),
'affiliation'    => get_bloginfo( 'name' ),
'value'          => $order->get_total(),
'tax'            => $order->get_total_tax(),
'shipping'       => $order->get_total_shipping(),
'currency'       => $order->get_currency(),
'items'          => $event_items,
``` 

Here we add them back to the details we send with the purchase event.

Closes #387

1. Setup the extension for tracking 
2. Create some products to add to the cart
3. Complete a purchase in an incognito window
4. Check the purchase event and confirm it contains both the correct transaction ID set to order ID and affiliation set to the site name (if shipping / taxes is enabled those totals can be confirmed as well)

Note: The changed E2E test still relies on a fix in #384 so it's a little harder to test.

### Additional details:
In the tests I didn't add a check for confirming the shipping totals and tax totals are tracked correctly. In order to do so we would need to set up a default shipping method. I think we can leave this for a follow up issue.

### Changelog entry
* Fix - Track purchase transaction ID.

### DIFF
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -178,10 +178,14 @@ export const purchase = ( { order } ) => {
 	}
 
 	return {
-		currency: order.totals.currency_code,
-		value: formatPrice(
-			order.totals.total_price,
-			order.totals.currency_minor_unit
+		transaction_id: order.id,
+		affiliation: order.affiliation,
+		currency: order.currency_code,
+		value: formatPrice( order.total_price, order.currency_minor_unit ),
+		tax: formatPrice( order.total_tax, order.currency_minor_unit ),
+		shipping: formatPrice(
+			order.total_shipping,
+			order.currency_minor_unit
 		),
 		items: order.items.map( getProductFieldObject ),
 	};
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -180,12 +180,18 @@ export const purchase = ( { order } ) => {
 	return {
 		transaction_id: order.id,
 		affiliation: order.affiliation,
-		currency: order.currency_code,
-		value: formatPrice( order.total_price, order.currency_minor_unit ),
-		tax: formatPrice( order.tax_total, order.currency_minor_unit ),
+		currency: order.totals.currency_code,
+		value: formatPrice(
+			order.totals.total_price,
+			order.totals.currency_minor_unit
+		),
+		tax: formatPrice(
+			order.totals.tax_total,
+			order.totals.currency_minor_unit
+		),
 		shipping: formatPrice(
-			order.shipping_total,
-			order.currency_minor_unit
+			order.totals.shipping_total,
+			order.totals.currency_minor_unit
 		),
 		items: order.items.map( getProductFieldObject ),
 	};
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -182,9 +182,9 @@ export const purchase = ( { order } ) => {
 		affiliation: order.affiliation,
 		currency: order.currency_code,
 		value: formatPrice( order.total_price, order.currency_minor_unit ),
-		tax: formatPrice( order.total_tax, order.currency_minor_unit ),
+		tax: formatPrice( order.tax_total, order.currency_minor_unit ),
 		shipping: formatPrice(
-			order.total_shipping,
+			order.shipping_total,
 			order.currency_minor_unit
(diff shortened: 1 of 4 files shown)


assets/js/src/tracker/data-formatting.js
```diff
@@ -178,10 +178,14 @@ export const purchase = ( { order } ) => {
...
-			order.totals.total_price,
-			order.totals.currency_minor_unit
+		transaction_id: order.id,
+		affiliation: order.affiliation,
+		currency: order.currency_code,
+		value: formatPrice( order.total_price, order.currency_minor_unit ),
+		tax: formatPrice( order.tax_total, order.currency_minor_unit ),
+		shipping: formatPrice(
+			order.shipping_total,
+			order.currency_minor_unit
```
tomalec: Maybe we could keep the same structure as we have for cart, and close to what WC Core uses:
```suggestion
		currency: order.totals.currency_code,
		value: formatPrice(
			order.totals.total_price,
			order.totals.currency_minor_unit
		),
		tax: formatPrice(
			order.totals.tax_total,
			order.totals.currency_minor_unit
		),
		shipping: formatPrice(
			order.totals.shipping_total,
			order.totals.currency_minor_unit
```

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -276,12 +276,14 @@ public function get_formatted_order( $order ): array {
...
-			),
-			'items'  => array_map(
+			'id'                  => $order->get_id(),
+			'affiliation'         => get_bloginfo( 'name' ),
+			'currency_code'       => $order->get_currency(),
+			'currency_minor_unit' => wc_get_price_decimals(),
+			'tax_total'           => $this->get_formatted_price( $order->get_total_tax() ),
+			'shipping_total'      => $this->get_formatted_price( $order->get_total_shipping() ),
+			'total_price'         => $this->get_formatted_price( $order->get_total() ),
+			'items'               => array_map(
```
tomalec: Same as https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/388/files#r1520324997

Maybe we should keep the same structure as before, as in cart, close to WC Core
```suggestion
			'id'          => $order->get_id(),
			'affiliation' => get_bloginfo( 'name' ),
			'totals'      => array(
				'currency_code'       => $order->get_currency(),
				'currency_minor_unit' => wc_get_price_decimals(),
				'tax_total'           => $this->get_formatted_price( $order->get_total_tax() ),
				'shipping_total'      => $this->get_formatted_price( $order->get_total_shipping() ),
				'total_price'         => $this->get_formatted_price( $order->get_total() ),
			),
			'items'        => array_map(
```

tomalec (approved):
Reviewed the code, tested locally, LGTM

The only non blocking comment I have is in regards to `totals` placeholder. @puntope, do you have an opinion here?

includes/class-wc-abstract-google-analytics-js.php
mikkamp: Thanks, that makes perfect sense. I was struggling with where to place the totals since I didn't want to repeat currency_code and currency_minor_unit for three different totals. But this is a nice way to combine it.

mikkamp:
Went ahead and implemented the suggestions for totals. Merging this now.

Issue #248, opened 2023-01-22
Expand tests coverage


martynmjones:
## Describe the issue
As part of the recent efforts to improve consistency across event data structures, https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/245 to avoid introducing issues in the future.

Now that there is a framework in place for unit testing the backend we should look to expand test coverage for the extension.

#### E2E tests
In #385 we added coverage for E2E tests, as a follow up to that PR we have the following list of things which aren't being tested:

- [ ] Remove from cart with shortcode `[woocommerce_cart]` isn't always tracked because of immediate page redirect (not in E2E)
- [ ] Remove from cart / Begin checkout is not consistent between blocks and shortcode, category and variation data differ.
- [ ] Change cart quantity is not tracked with shortcode `[woocommerce_cart]` or with blocks (PR #375 removes it)
- [ ] Add shipping info doesn't seem to be used anywhere (mentioned in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/375#issuecomment-1987887419)
- [ ] Select content is difficult to test if it is followed up by a page redirect (no E2E tests written yet)
- [ ] `begin_checkout` logs total item price (including discounts and price * quantity), `purchase` logs product object price (not final price we paid in the order). Should these be consistent?
- [ ] When we hook into `woocommerce_add_to_cart`, should we get final price from cart_item_data? Otherwise we'll track product prices vs. discounted price.
- [ ] When we change the settings to redirect to cart page after `add_to_cart`, do we still capture all details?
- [ ] Testing with grouped / bundled products (grouped product seems to only track first product added to cart)
- [ ] Shipping total and Tax total in purchase event
- [ ] Tests for the All Products block (no longer used for the default shop page)

PR #397, opened 2024-03-13
E2E tests for the All Products block

mikkamp:

This PR adds two additional E2E tests for the "All Products" block. Initially this block was used for the default Shop page, but that depends on which WooCommerce version is used. With the current version of WooCommerce the Shop page defaults to using the "Products (Beta)" block. So this PR ensures we explicitly test the block type regardless of which block is used for the shop page.

Like the other block page tests this currently tests two things:
1. Adds a product to the cart from the All Products block page
2. Checks that the event `view_item_list` is sent from the All Products block page

> [!Note]
The All Products block does not set a data attribute for each product so we can't select a specific add to cart button. However since the results are sorted by latest first we can pick the first add to cart button from the page.

This resolves one of the testing points in #248

1. Install all npm dependencies `nvm use && npm ci`
2. Build assets either through `npm start` or `npm run build`
3. Start the wp-env site npm run wp-env:up
4. Install browser (only needed once) `npx playwright install chromium`
5. Run tests npm run test:e2e
6. Confirm tests are successful (in particular the added All Products block tests)

### Changelog entry
* Dev - E2E tests for the All Products block.


martynmjones (approved):
Hey @mikkamp, thanks for the additional E2E tests!

Locally I'm getting some failures although they are not related to the `All Products` block. These tests look good to me so I've approved the PR and will open issues if needed tomorrow after investigating the failures more.

PR #398, opened 2024-03-13
Separate the global site tag setup and event tracking

martynmjones:

The changes to how the tracker is initiated in version `2.0.0` has caused problems with more sites than anticipated. Primarily those problems come from plugins that minify and combine JS files as many of them disregard listed dependencies while doing so.

As it currently stands there are many optimisation plugins that do work well with Google Analytics for WooCommerce but to make the extension more resilient against unexpected scenarios from different optimisation techniques this PR restructures how things execute.

- `dataLayer` and `gtag` (_or a function using the filtered tracker var name_) are no longer included in the JS file. Instead they are added directly to the page `<head>`
    - This is closer to how it was functioning pre-2.0.0
    - Does not use the WordPress script queue system to prevent this block of code being moved to anywhere else
    - Hooks into `wp_head` early so that the script is added as soon as possible. This makes it accessible to other plugins and means we are able to rely on it being available on the page regardless of when our scripts load 582e41b
- Relies on the `DOMContentLoaded` event to ensure inline script data is available at the point we begin tracking events 
    - E2E tests have been added to cover some of the scenarios we've seen with optimisation plugins that has caused problems with this 8e7481d
- Restructures the tracker to be more functional 9fede11

1. Build extension from `update/separate-site-tag-setup-and-event-tracking`
2. Run [smoke tests](https://github.com/woocommerce/woocommerce-google-analytics-integration/wiki/General-Testing#for-compatibility-and-general-smoke-testing) to confirm all events are tracked
3. Install an optimisation plugin like [Autoptimize](https://en-gb.wordpress.org/plugins/autoptimize/)
    - Configure the extension to combine and minify all Javascript
4. Check the page source to confirm that `gtag` is still setup in the page `<head>`
5. Re-run smoke tests to confirm events are still tracked

### Additional details:

- An adjustment was made to the selector for the related products add to cart E2E test as it was occasionally failing if a variable product was the first displayed in the related products list 25bd09c

### Changelog entry

> Update - Separate the site tag from the event tracking file and delay execution until DOMContentLoaded


includes/class-wc-google-gtag-js.php
```diff
@@ -214,29 +241,19 @@ public static function tracker_function_name(): string {
...
+			'linker'               => array(
 				'domains'        => ! empty( self::get( 'ga_linker_cross_domains' ) ) ? array_map( 'esc_js', explode( ',', self::get( 'ga_linker_cross_domains' ) ) ) : array(),
 				'allow_incoming' => 'yes' === self::get( 'ga_linker_allow_incoming_enabled' ),
 			),
-			'custom_map'            => array(
+			'custom_map'           => array(
 				'dimension1' => 'logged_in',
 			),
-			'events'                => self::get_enabled_events(),
-			'identifier'            => self::get( 'ga_product_identifier' ),
```
tomalec, line 244: :reminder_ribbon: We need to migrate that setting as well.

tomalec:
I have quite a revolutionary proposal, to change the `Tracker` from singleton to just a function and remove the config file: https://github.com/woocommerce/woocommerce-google-analytics-integration/compare/update/separate-site-tag-setup-and-event-tracking...update/separate-site-tag-setup-and-event-tracking-new-tracker?expand=1#diff-a745b91bc9c87614ca4293b765bb7a9053681512cb2cce7d9158b6e2946048b8R95

quite revolutionary, but TBH, at the stage of `2.0.0`, I think separating tracker and postponing blocks & classic tracking after `dom content loaded` could be disrupting anyway.

WDYT?

includes/class-wc-google-gtag-js.php
```diff
@@ -49,10 +49,40 @@ public function __construct( $settings = array() ) {
...
 	}
 
+	/**
+	 * Setup the global site tag as early as possible on the page
+	 *
+	 * @return void
+	 */
+	public function setup_site_tag() {
+		// phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedScript
+		printf(
```
tomalec: I wonder if we should be so brutal to break WP rules and print script directly; maybe enqueueing it as an inline with/ no dependencies would be enough? Or does the "optimization" plugins could force scripts enqueued in the head to be loaded later?

includes/class-wc-google-gtag-js.php
```diff
@@ -112,7 +140,7 @@ public function inline_script_data(): void {
 		wp_add_inline_script(
 			$this->data_script_handle,
 			sprintf(
-				'wcgai.trackClassicPages( %s );',
+				'var ga4wData = %s;',
```
tomalec: :nail_care: 
WDYT of separating 
```
				'var ga4w = {settings: %1$s, data: %2$s};',
```

tomalec:
If we aim to support "optimization" plugins,  it seems there is another plugin that turns parges into static ones, which undermines the entire idea of setting any data in PHP. See https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/399

includes/class-wc-google-gtag-js.php
martynmjones: Some of those plugins have settings to combine all inline JS and some will move them to a different file. As we know the script needs to be available at this point no matter what I think it makes sense that we avoid that possibility.

Instead of outputting it directly I've wrapped it in `wp_print_inline_script_tag` so that's we're still doing things the WordPress way 582e41b

tomalec:
> Does not use the WordPress script queue system to prevent this block of code from being moved to anywhere else

I wonder if we really need to go that extreme. Pre 2.0.0 implementation [used `wp_add_inline_script`]( https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/322/files#diff-75fada6b677c39f17a7de173a315db4d8911e7469f5a70dce4822a6aaa849514L323) and I assume it was working ok, with other extensions. Or maybe you have found an extension that breaks when we put dependency-less inline script using WP queue system?

I'm concerned that if we were using the WP queue before, then maybe there are extensions in the wild that rely on that. For example, there may be a need to use the WP queue system to hook their scripts before/after to add/overwrite some `gtag` config with precise timing.

tomalec (approved):
Tested locally with and w/o Autooptimizer, and reviewed the code. LGTM
I'd appreciate someone reviewing my code :)

assets/js/src/index.js
```diff
@@ -1,6 +1,24 @@
...
-import { trackClassicPages } from './integrations/classic';
-window.wcgai.trackClassicPages = trackClassicPages;
+import { createTracker } from './tracker';
+import { classicTracking } from './integrations/classic';
+import { blocksTracking } from './integrations/blocks';
 
-// Initialize tracking for Block based WooCommerce pages
-import './integrations/blocks';
+// Wait for DOMContentLoaded to make sure event data is in place.
+if ( document.readyState === 'loading' ) {
```
mikkamp: I don't know how to setup a test for this, but [according to this doc](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) there is also state `interactive`.

If the inline script for data (window.ga4w) is also optimized and delayed in loading then wouldn't the chance be bigger that we throw the error during this `interactive` stage.

I was under the impression that since we moved the first gtag configuration (plus consent mode) to the header, we'd be ok with setting up the actual tracking of events much later. Is there still a particular reason we can't use the `complete` state in combination with the `load` event?

assets/js/src/index.js
```diff
@@ -1,6 +1,24 @@
...
+} else {
+	eventuallyInitializeTracking();
+}
+function eventuallyInitializeTracking() {
+	if ( ! window.ga4w ) {
+		throw new Error(
+			'Google Analytics for WooCommerce: Configuration and tracking data not found.'
+		);
+	}
+	const eventHandler = createTracker( window.ga4w.settings );
```
mikkamp: Just a minor point of confusion, from my impression all that createTracker does is setup an eventHandler, does createEventHandler make more sense?

Also I can understand the simplicity of passing the list of enabled and tracker function name in the constructor. But it doesn't actually use them till an event is handled (which can be much later). I guess in this case that's not useful to change because the classicTracking does rely on the data being available on construction.

includes/class-wc-google-gtag-js.php
```diff
@@ -49,13 +49,40 @@ public function __construct( $settings = array() ) {
...
+					%2$s( "consent", "default", mode );
+				}
+				%2$s("js", new Date());
+				%2$s("set", "developer_id.%3$s", true);
+				%2$s("config", "%1$s", %5$s);',
+				esc_js( self::get( 'ga_id' ) ),
+				esc_js( self::tracker_function_name() ),
+				esc_js( self::DEVELOPER_ID ),
+				json_encode( self::get_consent_modes() ),
+				json_encode( $this->get_site_tag_config() )
```
mikkamp: Are we tied to legacy function names / visibility here? Just looks confusing to be mixing `self` and `$this`.

includes/class-wc-google-gtag-js.php
```diff
@@ -49,13 +49,40 @@ public function __construct( $settings = array() ) {
...
-	 * Register tracker scripts and its inline config.
-	 * We need to execute tracker.js w/ `gtag` configuration before any trackable action may happen.
+	 * Setup the global site tag as early as possible on the page
+	 *
+	 * @return void
+	 */
+	public function setup_site_tag() {
+		wp_print_inline_script_tag(
+			sprintf(
+				'/** Google Analytics for WooCommerce (gtag.js) */
```
mikkamp: If this is just a regular JS comment then we can use a regular `/*` comment or a `//`.

The extra indentation here also looks nice for the PHP code, but the final snippet doesn't look as clean.

includes/class-wc-google-gtag-js.php
```diff
@@ -49,13 +49,40 @@ public function __construct( $settings = array() ) {
...
 
 	/**
-	 * Register tracker scripts and its inline config.
-	 * We need to execute tracker.js w/ `gtag` configuration before any trackable action may happen.
+	 * Setup the global site tag as early as possible on the page
+	 *
+	 * @return void
+	 */
+	public function setup_site_tag() {
+		wp_print_inline_script_tag(
```
mikkamp: Does printing the snippet directly have any consequences for other plugins extending the code?

These solutions are not as compatible anymore: https://github.com/woocommerce/google-listings-and-ads/blob/2.6.4/src/Google/GlobalSiteTag.php#L220-L237

In theory the filter `woocommerce-google-analytics-integration` still works, but that's added much later and not right after the snippet in the head (the timing when we want to set configuration).

We can't use the filter `woocommerce_ga_gtag_config` because we don't want to add additional config parameters, we want to add tracking to a second GA_ID. Maybe moving the three `gtag` lines to an array of strings and then filtering the array before each line is added would be more flexible.

mikkamp (commented):
I left some additional comments as I looked through the code.

As @tomalec mentioned, I'm also a bit concerned about the use of `wp_print_inline_script_tag` compared to the previous use of `wp_add_inline_script`.

There is no filter for the contents of `wp_print_inline_script_tag` only for the script attributes. Whereas `wp_add_inline_script` allows us a lot more flexibility. That becomes evident with the GLA compatibility where it isn't able to set config data right after the snippet in the head.

assets/js/src/index.js
tomalec: >Just a minor point of confusion, from my impression all that createTracker does is setup an eventHandler, does createEventHandler make more sense?

I have hard time with thinking of the names ;) as we have
1. [the `gtag` (or renamed) function](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/files#diff-efaa51133b4f4fcb6d50a840b6c0e00dd53244078149890409ddbe43ce3d7408R32)  - "tracking function"?
2. [Function to be fed with event data](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/files#diff-efaa51133b4f4fcb6d50a840b6c0e00dd53244078149890409ddbe43ce3d7408R29), that formats it and calls 1. This function is given to Block action - the actual "event handler"?
3. [Function that takes event name](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/files#diff-efaa51133b4f4fcb6d50a840b6c0e00dd53244078149890409ddbe43ce3d7408R22), picks the formatted for it and produces a function 2. - "create/get event handler"?
4. [Function that makes the global configuration](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/files#diff-efaa51133b4f4fcb6d50a840b6c0e00dd53244078149890409ddbe43ce3d7408R11) to produce 3.
 - "create event handler creator"?
 
 So I think of a two sets of new names 
 1. "tracker" plus its `trackerFunctionName`
 2. `trackerEventHandler`
 3. `eventHandler`
 4. `createEventHandler`
 
 or 

 1. "tracker"
 2. `eventHandler`
 3. `getEventHandler`
 4. `createEventHandlers` - but technically, it does not create event handlers, it creates a function to create them.

or we could change it even more, as instead of creating a new function every time the function 3. is called. we could create all of them as once 4 is called and put it in a map, something like:

```js
const eventHandlers = careateEventHandlers( window.ga4w.settings ); // new EventHandlers(  window.ga4w.settings  );
classicTracking( eventHandler, window.ga4w.data );
blocksTracking( eventHandler );
// ...
addUniqueAction(`foo`, 'bar', eventHandlers.get( 'view_item' ) ); 
// or
addUniqueAction(`foo`, 'bar', eventHandlers[ 'view_item' ] );
// ...
eventHandlers.get( 'add_to_cart' )( { product: productToHandle } );
eventHandlers[ 'add_to_cart' ]( { product: productToHandle } );
```
Which is basically turning 3. to a more native, map-accessor function:
1. "tracker"
 2. `eventHandler`
 3. `Map.prototype.get` / `Object`'s property accessor
 4. `createEventHandlers`

tomalec: > But it doesn't actually use them till an event is handled (which can be much later). I guess in this case that's not useful to change because the `classicTracking` does rely on the data being available on construction.

I'm not sure if I get your point here.

My point was to pass the config once in the entry. Then, all the code that binds the event handlers to actions or DOM events would not have to keep passing it. Would not have to access it from the global scope or cascade it through all the parent functions.

mikkamp: > I'm not sure if I get your point here.

Sorry about the confusion there. I was analyzing the `createTracker` first and realized that it didn't need to throw the error `Google Analytics for WooCommerce: Configuration and tracking data not found.`, as it only needs the settings later. But then I inspected classicTracker and realized that there the data is needed right away. So in theory we could load createHandler without throwing the warning but it's not much use to do so as it's only used by the classicTracking and blockTracking. So to summarize the comment was my train of thought but not valid to apply.

mikkamp: > I have hard time with thinking of the names ;)

Naming is definitely a hard topic, but it was just a minor point, no use in doing that much refactoring for my sake.

assets/js/src/index.js
tomalec: I took this construct from [MDN example](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete) 

----

> If the inline script for data (window.ga4w) is also optimized and delayed in loading then wouldn't the chance be bigger that we throw the error during this `interactive` stage.

Do you mean the case when inline data is deferred to the `interactive` phase, while our JS may run before?

I made a few synthetic test cases to achieve such effect:
- wrapping the inline data inside `setTimeout` https://stackblitz.com/edit/inline-script-in-interactive-document-state?file=index.html
- wrapping inline data inside inline **async module** https://stackblitz.com/edit/inline-script-in-interactive-document-state-rcxnmd?file=index.html

But do you think we should support that?
1. If somebody is messing with the scripts by changing them from blocking to asynchronous, there are bigger problems to expect
3. I didn't test it yet, but I wonder if running on "load" may be too late if we want to catch some events, before page redirects, etc.


---

On the other hand, if we're going to prepare ourselves to be run async, we could as well deliberately set up ourselves using async modules. I didn't test the new API to this extend yet. Especially since there were some discussions regarding the lack of inline module support in WP.  But then we could do something like:

```html
<!-- WP-build import map -->
<script type="importmap">
  {
    "imports": {
      "woocommerce-google-analytics-integration": "path/to/main.js",
    }
  }
</script>
<script id="woocommerce-google-analytics-integration" type="module" src="path/to/main.js" async></script>
<script type="module" async>
import initializeTracking from 'woocommerce-google-analytics-integration';

const ga4w = { /* inlined JSON */}
initializeTracking( ga4w );
</script>
```


But still, what "optimizations" for such scripts we should anticipate, I don't know.

includes/class-wc-google-gtag-js.php
tomalec: fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/commits/4d6f371edce123fcfd2be312e496a43b1376b503

assets/js/src/index.js
tomalec: Speaking of function names, WDYT of https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/commits/cdb60c08a9cb62f351f5d324472ad62b6eef7105 ?

includes/class-wc-google-gtag-js.php
tomalec: I had a bigger concern with the `self::` functions, especially `self::get`. 
But I didn't want to push it here, not to refactor everything at once.
We use a lot of `static` functions that are not 100% "static," IMO.

[_PHP spec_](https://www.php.net/manual/en/language.oop5.static.php)
> Declaring class properties or methods as static makes them accessible without needing an instantiation of the class.

- `public static function get_enabled_events` uses `self::get`
- `public static function load_opt_out` uses `self::get`
- `public static function get_product_identifier` uses `self::get`
- `public static function get` uses the `self::settings`
**but** we set the `self::settings` in the instance constructor https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/cdb60c08a9cb62f351f5d324472ad62b6eef7105/includes/class-wc-google-gtag-js.php#L47:L47

So we could change `self::get( * )` to `$this->get()` or event `$this->settings[ * ]` (IMHO, `get` just blurs the picture here w/o much benefit).


Do you also suggest changing `tracker_function_name`, `DEVELOPER_ID`, and `get_consent_modes` to be instance methods/props?

tomalec:
Thank you, @mikkamp, for the detailed review!
I addressed some of your comments. AFAIK, @martynmjones is testing whether there are some extreme optimization plugins that would cause problems with a regularly enqueued inline script with `gtag` definition.
In the meantime, I'd appreciate your answers in the remaining threads.

includes/class-wc-google-gtag-js.php
mikkamp: I'd suggest in this case to access the methods through the current instance, which will make it easier in the future to remove `static` from those function declarations. We could do that now but then code snippets / other plugins would have a little harder time getting the instance.

```suggestion
				esc_js( $this->get( 'ga_id' ) ),
				esc_js( $this->tracker_function_name() ),
				esc_js( self::DEVELOPER_ID ),
				json_encode( $this->get_consent_modes() ),
				json_encode( $this->get_site_tag_config() )
```

Generally the purpose of the `get` helper function would be to enforce both the parameter type for key and the return value. However it seems the return value for non_scalar values is broken. If you do something like this you'll get a fatal error:
```php
self::$settings['ga_id'] = [];
var_dump( $this->get( 'ga_id' ) );
```

Where we do see it working is for example if we do something like:
```php
self::$settings['ga_id'] = 1;
var_dump( $this->get( 'ga_id' ), self::$settings['ga_id'] );
```
The first value will be a string, but the second will be an int, so we will be less certain what kind of value we will be working with if we don't use the `get` method.

assets/js/src/index.js
mikkamp: I think that's a bit clearer. I don't see any need to refactor beyond that, as it will just be nitpicking.

assets/js/src/index.js
mikkamp: I was referring to some of the more aggressive options in plugins like Autoptimize:
image

I haven't actually tested all those scenarios, and regardless of how it's aggregated, I'd expect the load order to remain the same within the aggregated script. Although looking at the forum there was plenty cases where the load order was changed as well [as plugins](https://wordpress.org/plugins/async-javascript/) that purely allow you to set `async`, which means we can no longer guarantee load order.

> But do you think we should support that?
> 1. If somebody is messing with the scripts by changing them from blocking to asynchronous, there are bigger problems to expect
> 2. I didn't test it yet, but I wonder if running on "load" may be too late if we want to catch some events, before page redirects, etc.

I wouldn't expect us to be able to support every scenario, but as long as we are compatible with a handful of main optimizers and there are options that can be set to exclude particular scripts then I'd find that sufficient.

> may be too late if we want to catch some events, before page redirects

As far as I've seen the Google gtag script has a delay (at least in my cases a couple seconds) after load so page redirects that happen upon load would almost always be too late to catch.

I like the async module solution. It looks like a stable solution, although I haven't worked with that before, so unable to determine if that would have any secondary effects.

mikkamp (commented):
Thanks for addressing the additional comments, I've added some additional thoughts there.

At this point the only thing in this PR that I'm a little on the fence about is:

> There is no filter for the contents of `wp_print_inline_script_tag` only for the script attributes. Whereas `wp_add_inline_script` allows us a lot more flexibility. That becomes evident with the GLA compatibility where it isn't able to set config data right after the snippet in the head.

GLA still injects triggers for the events directly within the page body, so we'll want to ensure that the configuration is set early (in the head). One of the following options could resolve that:
1. Allow filtering of the snippet added in `wp_print_inline_script_tag` (the whole snippet, or filter for adding extra lines)
2. Revert back to using `wp_add_inline_script` so we can enqueue configuration right after the script.

In addition to that I think if we'd pick like 3 or 4 of the most popular optimizer plugins, if we are compatible with more or less the default configurations for optimizing JS then I'd see that as acceptable to release this PR.

includes/class-wc-google-gtag-js.php
tomalec: Thanks :) I changed the reference in non-static methods to use `$this->` in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/commits/867f69657d7a1e48713ab49ab0cb56d594ede9d7 (plus [tweaked](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/commits/89487e643cb47a2030be2de9160d3b011532dd18) `self::DEVELOPER_ID` to `static::DEVELOPER_ID`, to suppress the notice).

Let's change the `static` in methods in a separate PR and release. I'll start a draft to keep this idea on track

tomalec: > Generally the purpose of the get helper function would be to enforce both the parameter type for key and the return value. However it seems the return value for non_scalar values is broken. If you do something like this, you'll get a fatal error:

I wonder if we really need such behavior. I was experimenting with adding array-like settings for consent modes, and this stringification behavior was more of a problem than a help. But that's also something to address in a separate PR

assets/js/src/index.js
tomalec: Thanks for clarifications, I'll do more experiments with Autooptimize.

> I like the async module solution. It looks like a stable solution, although I haven't worked with that before, so unable to determine if that would have any secondary effects.

I did some experiments, but i think it's too early to do so:
- Script Modules API is only available in WP 6.5 (not yet released)
- it does not support inline modules - we could not pass the data that way
- modules cannot have script dependencies - we could not add the classic `google-tag-manager` script as a dependency

includes/class-wc-google-gtag-js.php
mikkamp: Definitely limiting, and assuming of what settings are like. Agree that it would be better handled in a separate PR.

assets/js/src/index.js
mikkamp: Sounds like modules are still a little limiting.
We need to always register a script with the `google-tag-manager` ID, as otherwise WC core will load it's fallback. So we'd have to keep that in mind for resolving dependencies. But as you mentioned it's not ready yet, so at most I'd add the information we currently know to an enhancement issue so it's easy to track.

martynmjones:
> In addition to that I think if we'd pick like 3 or 4 of the most popular optimizer plugins, if we are compatible with more or less the default configurations for optimizing JS then I'd see that as acceptable to release this PR.

I've run some tests with a few of the popular optimisation plugins using the different configurations on offer. There have been two issues with Autoptimize, `Aggregate inline JS` as you pointed out earlier, and forcing the JS script to load in the `<head>`.

If `main.js` is moved to the `<head>` then it will continue working and we have a [test to cover that scenario](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/89487e643cb47a2030be2de9160d3b011532dd18/tests/e2e/specs/js-scripts/load-order.test.js#L21). The problem with that setting in Autoptimize is that dependencies are ignored so several things break and not just Google Analytics for WooCommerce. So I don't think we need to make adjustments to try to have the extension work with that setting.

For `Aggregate inline JS`, the easiest fix is to add `ga4w` to the list of excluded scripts/ variables in Autoptimize. So an option would be to keep things as they are an add a "Known incompatibilities" section to the docs and explain to merchants who want to have inline JS aggregated how to exclude the script?

| Plugin Name | Minify JS | Combine JS Files | Aggregate inline JS | Defer JS | Delay Execution | Force JS in `<head>` |
| - | :-: | :-: | :-: | :-: | :-: | :-: |
| [Autoptimize](https://en-gb.wordpress.org/plugins/autoptimize/)  | ✅ | ✅ | ❌ | ✅ | n/a | ❌ |
| [WP-Optimize](https://en-gb.wordpress.org/plugins/wp-optimize/)  | ✅ | ✅ | n/a | ✅ | n/a | n/a |
| [WP Rocket](https://wp-rocket.me/)  | ✅ | ✅ | n/a | ✅ | ✅ | n/a |
| [W3 Total Cache](https://en-gb.wordpress.org/plugins/w3-total-cache/)  | ✅ | ✅ | ✅ | ✅ | ✅ | n/a |
| [LiteSpeed Cache](https://wordpress.org/plugins/litespeed-cache/)  | ✅ | ✅ | ✅ | ✅ | ✅ | n/a |

tomalec:
Thanks for the nice comparison @martynmjones !

I think we can try to make workarounds to make our code work if the scripts are aggregated, loaded in random order, or extracted to external JS. They question is: do we want to? 

Make JS work even if data is loaded from an external `defer`ed `async` script or an external file—but if someone puts that data into an external file, is it still dynamically populated with the data we collect with hooks?

We could assume it is. But if it's not it may silently corrupt the behavior, which will be hard to identify by HEs.

-----

Alternatively, we could make JS fully static, defer, and async by default, and put the data into HTML, like: 
`<wc-ga4w event-data="{JSON with data}">`
This will let all the optimizers optimize the JS, which could be cached, deferred, etc, but we'll keep our data safe in the markup.
And also would me it easy to pick up the timing when the data gets there: `customElements.define('wc-ga4w', class extends HTMLElement{ attributeChangedCallback });`

Then:
 - if JS is loaded before the element is stamped to the document, the element will do the job when it's ready
 - otherwise, the element will do the job once it's upgraded == when the JS gets loaded
 - our JS could have the `load` event listener to throw a warning if there is not `document.querySelector('wc-ga4w')` after `the document is `complete` => meaning the JS was loaded but expected data is missing.

Plus, we'll have the good old separation of concerns: JS is for behavior, HTML is for the content.

tomalec:
@martynmjones @mikkamp I made a change in our logic flow and conditions: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/commits/dacaabcf29aff66f310fd71f90cd5f335a7d5285
Now:
- `woocommerce-google-analytics-integration-data` sets the data global and dispatches an event to `document`
- `woocommerce-google-analytics-integration`:
   - Use the data if it's already there, to process it ASAP
   - If it's not there: 
      - Add an event listener for our own event to initialize once data is ready, 
      - Warn if/once the document is fully loaded, and the data is still missing.

I believe this way, we're more bulletproof towards other extensions messing up with our order and timing.

Pros:
 - Autoptimize "Aggregate inline JS" works
 - I'm not sure how about "Force JS in `<head>`" as for me, it breaks other parts of Woo, but theoretically, it should work as well, as long as both scripts eventually get somewhat loaded.
 - More lightweight than a custom element implementation.

Cons:
- Autoptimize "Aggregate inline JS" works; however, it puts dynamic, cart-specific data into a cacheable, external JS file, which would malform the data silently.
- "Pollutes" the document with one event
- Will keep showing warnings for folks who load `woocommerce-google-analytics-integration` in blocking fashion but `woocommerce-google-analytics-integration-data` - but maybe that's actually good thing
- It assumes `ga4w:ready` means `window.ga4w` is ours and ready. Will not protect us against somebody triggering `ga4w:ready` event on their own w/o `window.ga4w` available. Custom element way would have a bit more integrity in that regard. But I guess we cannot protect against everything.
- Puts a bit more JS in PHP string

assets/js/src/index.js
tomalec: I made an issue for ESModules approach https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/411

and tweaked the wait for data flow in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398/commits/dacaabcf29aff66f310fd71f90cd5f335a7d5285. So it no longer throws, but use the data once it comes. Which make GA4W kinda work with Autoptimize "Aggregate inline JS". But that option puts dynamic data into cached file, which is a separate problem.

includes/class-wc-google-gtag-js.php
tomalec: Added an issue to track that enhancement idea https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/412

assets/js/src/index.js
mikkamp: Thanks I like that solution.

mikkamp:
Thanks @martynmjones for all the testing, that's a great result, and I think sufficient enough to move forward.

@tomalec I like your last addition and I think that makes it a bit more robust.

> it puts dynamic, cart-specific data into a cacheable, external JS file, which would malform the data silently

Having the cart specific data here is not the most ideal solution. Instead we should be fetching it from the cart fragments or the data used to populate the mini cart, but that's getting a little out of scope and goes beyond what we had pre 2.0

As for the other cons you listed, I think it's good to be aware of them, but not necessarily something we want to protect against. Placing some expectations in the documentation would probably be a better way to go about it for now.

So from my perspective this PR is pretty much ready to go. I'm just not sure what the final conclusion to this comment was: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398#discussion_r1539319787
Are we still looking to make that more flexible here?

martynmjones:
> So from my perspective this PR is pretty much ready to go. I'm just not sure what the final conclusion to this comment was: https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398#discussion_r1539319787
Are we still looking to make that more flexible here?

Thanks for all the input @mikkamp 👍 I updated the PR so it's closer to how we were previously doing it 77067d8

PR #400, opened 2024-03-14
Fix warning after plugin install using wp-cli

steffans:
Fixes PHP Warning after plugin install using wp-cli.

`Undefined array key "ga_id" in /wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-google-analytics.php on line 269`

### Changelog entry

> Fix - Warning after plugin install using wp-cli.

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -266,7 +266,7 @@ class WC_Google_Analytics extends WC_Integration {
 	 * @return bool         True if tracking for a certain setting is disabled
 	 */
 	private function disable_tracking( $type ) {
-		return is_admin() || current_user_can( 'manage_options' ) || ( ! $this->settings['ga_id'] ) || 'no' === $type || apply_filters( 'woocommerce_ga_disable_tracking', false, $type );
+		return is_admin() || current_user_can( 'manage_options' ) || empty( $this->settings['ga_id'] ) || 'no' === $type || apply_filters( 'woocommerce_ga_disable_tracking', false, $type );
 	}
 
 	/**


tomalec:
Hi @steffans, thank you for your contribution! :) 

The code itself looks good, but I was not able to reproduce the warning you mentioned.
For me the [`init_settings` in the constructor](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/36733d47b45f9b18e1ebf79fbeadc813b016360a/includes/class-wc-google-analytics.php#L39:L39) successfully creates the `->settings` array full of defaults.

Could you elaborate more on the steps to reproduce?

I'd like to make sure there is no other bug hidden somewhere else :)

tomalec:
I managed to reproduce the issue:
1. Deactivate, uninstall GA4W.
2. Remove `woocommerce_google_analytics_settings` entry from `wp_options`
3. Install and activate using CLI
```sh
wp plugin install woocommerce-google-analytics-integration
wp plugin activate  woocommerce-google-analytics-integration
``` 
4. Visit any page as a guest.

tomalec:
The reported issue from OP is also the reason why our QIT activation fails.

It seems that for such activation, the `init_settings` fails to set default values, as it finds the options already filled and does nothing.

The options are set by [`maybe_set_defaults`](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/trunk/woocommerce-google-analytics-integration.php#L226:L231) but it sets only `ga_product_identifier`
I have 2 doubts:
1. Do we even need `maybe_set_defaults`? Isn't the `init_settings` enough for use to get the defaults?
2. If we need it, shoulnd't we set all the other settings to their respective defaults too?

@martynmjones do you have some ideas?

tomalec (approved):
I reviewed the code, and tested it locally LGTM
The change fixes CLI activation, as well as running the extension with old, outdated, or malformed options.

steffans:
I noticed this issue when installing multiple plugins in a clean Wordpress using the wp-cli:

```
wp plugin install woocommerce-google-analytics-integration --activate
wp plugin install unbloater --activate

Installing Unbloater (1.6.1)
Downloading installation package from https://downloads.wordpress.org/plugin/unbloater.zip...
Unpacking the package...
Installing the plugin...
Removing the old version of the plugin...
Plugin updated successfully.
Activating 'unbloater'...
Plugin 'unbloater' activated.
Success: Installed 1 of 1 plugins.
Warning: Undefined array key "ga_id" in /wp-content/plugins/woocommerce-google-analytics-integration/includes/class-wc-google-analytics.php on line 269
```

martynmjones:
@tomalec thanks for the ping.

> The options are set by [maybe_set_defaults](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/trunk/woocommerce-google-analytics-integration.php#L226:L231) but it sets only ga_product_identifier
I have 2 doubts:
>
> 1. Do we even need maybe_set_defaults? Isn't the init_settings enough for use to get the defaults?
> 2. If we need it, shoulnd't we set all the other settings to their respective defaults too?

In version `2.0.0` a setting was added for `Product Identification` which we want to default to product ID for new merchants and product SKU for existing merchants.

`maybe_set_defaults` only sets the product identification option to `Product ID` during activation if the extension has not yet been configured on that website. If we were to remove that function and change the current default value specified in the [form fields array](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/7a04842cfae7ad5001f41eba10280f98084ed97d/includes/class-wc-google-analytics.php#L107) then products would start to be identified differently to Google for existing merchants when they upgrade to version `2.0.0` or above which we need to avoid.

So eventually we may be able to remove `maybe_set_defaults` and revisit how we're working for settings but for now it's still needed.

tomalec:
Thanks for the clarifications @martynmjones 
So it seems we should do the opposite, set all the other defaults together with `Product Identification`, as the current `maybe_set_defaults` breaks the `init_settings` behavior. I created a follow-up issue for that https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/408

tomalec (approved):

PR #402, opened 2024-03-15
Add require plugins header

### LINKED ISSUE #401: Add new plugin dependency headers
Since WP 6.5 and In relation to this post: pcShBQ-1yd-p2

There is also the suggestion to add the  new header:

```
Requires Plugins: woocommerce, automatewoo
```

This header prevents the plugin from being installed or activated when the header is present. 

See also https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/ to learn about the new dependency header and what to expect about it.

puntope:

Closes #401   

This PR enables the new Requires Plugins header available in WP 6.5

See --> https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/

### Screenshots:

Screenshot 2024-03-15 at 10 53 45

Screenshot 2024-03-15 at 10 53 35

1. Install WP 6.5 (RC2) --> https://wordpress.org/wordpress-6.5-RC2.zip 
2. Install Woo Commerce and the extension (checkout this PR)
3. If WC is activated, the extension can be activated
4. If WC is not activated, extension button for activate is disabled. A notice "This plugin cannot be activated because required plugins are missing or inactive." is shown.
5. If extension is activated and WC is activated. WC cannot be deactivated. And it shows info like "Required by: Google Analytics for WooCommerce" 

### Changelog entry

> Tweak - Add WP 6.5 Require plugins header

### DIFF
--- a/woocommerce-google-analytics-integration.php
+++ b/woocommerce-google-analytics-integration.php
@@ -9,6 +9,7 @@
  * WC requires at least: 8.4
  * WC tested up to: 8.7
  * Requires at least: 6.2
+ * Requires Plugins: woocommerce
  * Tested up to: 6.5
  * License: GPLv2 or later
  * Text Domain: woocommerce-google-analytics-integration


martynmjones (approved):
Hey @puntope, thanks for the updates!

Confirmed that the extension can't be activated without WooCommerce installed and get the expected messaging when they are both installed so LGTM ✅

PR #419, opened 2024-05-09
Apply discounts to the ecommerce event if available

### LINKED ISSUE #417: Item Revenue data sent to Google Analytics uses Product Price, not Order Item Total
### Describe the bug:
There is a discrepancy between the Revenue reported in WooCommerce and the Total Item Revenue shown in Google Analytics 4 (GA4) under "Monetization" >> "Ecommerce purchases". The discrepancy originates from the plugin sending the regular price of a product to Google Analytics, rather than the item total as displayed in the order. For example, in a recent order, the customer purchased an art kit priced at $125 and added an online art course regularly priced at $45. However, since the customer used a coupon code, the item total for the course was $0:

Screen Shot 2024-05-07 at 11 39 51 PM

Yet, in GA4, the course appears with its regular price, leading to a discrepancy in total item revenue reporting.

Screen Shot 2024-05-07 at 11 52 58 PM
Screen Shot 2024-05-07 at 11 41 23 PM

### Steps to reproduce:

1. Complete an order with multiple items where at least one item is discounted from the regular price through a coupon or other promotional adjustments.
2. Check the item revenue in GA4 and compare it with the WooCommerce order summary.

### Expected behavior:
The item revenue in GA4 should reflect the actual amount paid as shown in the WooCommerce order summary, accounting for any discounts applied.

### Actual behavior:
The plugin is incorrectly sending the regular product price to GA4, rather than the discounted total shown on the order. This results in inflated revenue figures in GA4 that do not match the actual store revenue.

ianlin:

Closes #417. This PR fixes a bug that the price in purchase events does not take coupon discounts into account by getting the price after discount using [the method `get_total()` in `WC_Order_Item_Product` class](https://github.com/woocommerce/woocommerce/blob/54eba223b8dec015c91a13423f9eced09e96f399/plugins/woocommerce/includes/class-wc-order-item-product.php#L308-L310), and [add the `discount` parameter to the ecommerce event](https://developers.google.com/analytics/devguides/collection/ga4/apply-discount?client_type=gtag).

Only the `purchase` event will include discounts if available, since only the method `get_formatted_order()` will include the "price after coupon discounts" for each order line item.

### Checks:

1. Checkout the branch of this PR and run `npm ci` and `npm run build`
2. Open Network tab from browser dev console, add `google-analytics` as the filter
3. Add some products into the cart, for example, two products.
4. Go to checkout page
5. Apply a coupon
6. Complete the purchase, see the order received page
7. From the Network tab, find the request with `dt: Order Confirmation` and `en: purchase`, which is the purchase event
8. From this event you should see there are two product parameters starts with `pr1` and `pr2`, e.g.
    - **pr1**: `id28~nmSingle~caMusic~qt1~pr12~ds12`
    - **pr2**: `id36~nmBeanie with Logo~caAccessories~qt1~pr10~ds8`
9. The `pr` in the value is the price after coupon discounts, and `ds` is the amount of coupon discounts.
10. Add another two products into the carts, complete the order WITHOUT applying any coupons.
11. You should see there is no `ds` in the event parameters, and the price is just the regular price.

### Changelog entry

> Fix - Apply discounts to the ecommerce event if available

### DIFF
--- a/assets/js/src/utils/index.js
+++ b/assets/js/src/utils/index.js
@@ -15,7 +15,7 @@ export const getProductFieldObject = ( product, quantity ) => {
 		variantData.item_variant = product.variation;
 	}
 
-	return {
+	const data = {
 		item_id: getProductId( product ),
 		item_name: product.name,
 		...getProductCategories( product ),
@@ -26,6 +26,23 @@ export const getProductFieldObject = ( product, quantity ) => {
 		),
 		...variantData,
 	};
+
+	// Apply discounts to ecommerce events.
+	// https://developers.google.com/analytics/devguides/collection/ga4/apply-discount?client_type=gtag
+	if (
+		typeof product.price_after_coupon_discount === 'number' &&
+		! isNaN( product.price_after_coupon_discount ) &&
+		product.price_after_coupon_discount !== product.prices.price
+	) {
+		data.discount =
+			product.prices.price - product.price_after_coupon_discount;
+		data.price = formatPrice(
+			product.price_after_coupon_discount,
+			product.prices.currency_minor_unit
+		);
+	}
+
+	return data;
 };
 
 /**
--- a/assets/js/src/utils/index.js
+++ b/assets/js/src/utils/index.js
@@ -32,7 +32,7 @@ export const getProductFieldObject = ( product, quantity ) => {
 	if (
 		typeof product.price_after_coupon_discount === 'number' &&
 		! isNaN( product.price_after_coupon_discount ) &&
-		product.price_after_coupon_discount !== product.prices.price
+		product.price_after_coupon_discount < product.prices.price
 	) {
 		data.discount = formatPrice(
 			product.prices.price - product.price_after_coupon_discount,
--- a/assets/js/src/utils/index.js
+++ b/assets/js/src/utils/index.js
@@ -34,8 +34,10 @@ export const getProductFieldObject = ( product, quantity ) => {
 		! isNaN( product.price_after_coupon_discount ) &&
 		product.price_after_coupon_discount !== product.prices.price
 	) {
-		data.discount =
-			product.prices.price - product.price_after_coupon_discount;
+		data.discount = formatPrice(
+			product.prices.price - product.price_after_coupon_discount,
+			product.prices.currency_minor_unit
+		);
 		data.price = formatPrice(
 			product.price_after_coupon_discount,
 			product.prices.currency_minor_unit
--- a/assets/js/src/utils/index.js
+++ b/assets/js/src/utils/index.js
@@ -29,11 +29,7 @@ export const getProductFieldObject = ( product, quantity ) => {
 
 	// Apply discounts to ecommerce events.
 	// https://developers.google.com/analytics/devguides/collection/ga4/apply-discount?client_type=gtag
-	if (
-		typeof product.price_after_coupon_discount === 'number' &&
-		! isNaN( product.price_after_coupon_discount ) &&
-		product.price_after_coupon_discount < product.prices.price
-	) {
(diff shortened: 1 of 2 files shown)


assets/js/src/utils/index.js
```diff
@@ -26,6 +26,23 @@ export const getProductFieldObject = ( product, quantity ) => {
 		),
 		...variantData,
 	};
+
+	// Apply discounts to ecommerce events.
+	// https://developers.google.com/analytics/devguides/collection/ga4/apply-discount?client_type=gtag
+	if (
+		typeof product.price_after_coupon_discount === 'number' &&
+		! isNaN( product.price_after_coupon_discount ) &&
```
puntope: Why do we need these two? `product.price_after_coupon_discount` seems to be always a number

assets/js/src/utils/index.js
```diff
@@ -26,6 +26,23 @@ export const getProductFieldObject = ( product, quantity ) => {
...
+		product.price_after_coupon_discount !== product.prices.price
```
puntope: WDYT about changing this to `product.price_after_coupon_discount < product.prices.price` I don't think we can have negative discounts.

assets/js/src/utils/index.js
```diff
@@ -26,6 +26,23 @@ export const getProductFieldObject = ( product, quantity ) => {
...
+	) {
+		data.discount =
+			product.prices.price - product.price_after_coupon_discount;
```
puntope, line 38: We need to wrap this with formatPrice() as well. Otherwise the discount is send without decimals.

puntope: We need to wrap this with formatPrice() as well. Otherwise, the discount is sent without decimals. 

Screenshot 2024-05-10 at 16 25 13

puntope (commented):
Thanks for working on this @ianlin 

The price is shown correctly but not the discount.

See my comments in the PR. Thanks

assets/js/src/utils/index.js
ianlin: Since there were not only purchase event would call the method `getProductFieldObject()`, for example `begin_checkout` event would call this method as well. Initially I wanted to test if `product.price_after_coupon_discount` is defined but since the price after discount could be `0` and `0` is treated as `false`, that's why I added these two conditions here.

assets/js/src/utils/index.js
ianlin: Thanks for the nice catch!

ianlin:
Hey @puntope, thanks for the review and nice suggestions. I've pushed more commits and it's ready for another round.

assets/js/src/utils/index.js
puntope: Ah, I see. Thanks for clarifying. That might explain why we used

`if ( typeof product.price_after_coupon_discount === 'number' )`

But what about the NaN line check? 

Also WDYT of doing something like:

`if ( product?.price_after_coupon_discount < product.prices.price  ) `

So we can save both checks number and nan 

You can test it but in case `product.price_after_coupon_discount` is undefined this would be false still.

puntope (approved):
Thanks for the changes @ianlin 

I left one more comment regarding the need for the number and nan check. But approved in advance

assets/js/src/utils/index.js
ianlin: @puntope 

> But what about the NaN line check?

I added it because I see `typeof NaN` is also a `number`, so added it for some prevention.

> Also WDYT of doing something like:
if ( product?.price_after_coupon_discount < product.prices.price  ) 
So we can save both checks number and nan
You can test it but in case product.price_after_coupon_discount is undefined this would be false still.

Ah good point, thanks for the info.

PR #421, opened 2024-05-13
Fix incorrect quantity when product is already in cart

### LINKED ISSUE #391: Cart quantity can be inconsistent
### Describe the bug:
This is a followup from #386 

We search for products in the cart items first and then the list of products, [see here](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/2.0.1/assets/js/src/utils/index.js#L167-L168).

That's a problem when it comes to the quantity number, which we can see in the following scenario:
1. Add single product 3x to cart
2. Refresh archive page
3. Add same single product to cart
4. View tracking data and the quantity is set to 3

Ideally we should be setting the quantity back to 1. Also is it really relevant to search through cart data, even though the product ID matches, we might have added one with a discount / addon which could make the price different. We'll need to log this in a separate issue.

### Steps to reproduce:
1. Setup extension for tracking add to cart events 
2. Create a simple product
3. Create a page with the `[products]` shortcode
4. In an incognito window visit the archive page
5. Add the single product to the cart 3 times
6. Confirm each event is tracked for a single quantity being added
7. Refresh the archive page
8. Add the same single product to the cart
9. Confirm the event now has a quantity of 3 (even though we are adding the 4th item)

### Expected behavior:
The quantity always to be tracking the amount we are adding to the cart
The product price to be reflective of the cart item we are adding, and not a previously stored value of the product price

martynmjones:

Closes #391 

Adjusts the utility to retrieve a [product by its ID](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/421/commits/aac5606b817b8c1c46f7d80ed65ac59a12d0e33f) to prioritise the product data object over the cart object. This is to prevent the existing cart quantity from being sent with an `add_to_cart` event.

Additionally, the function to [format product data](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/421/commits/523932fb711428ae4bd56830ca59ba91b70ee3a4) is updated to pass along the quantity on the `added_to_cart` event.

1. Checkout `fix/391-incorrect-product-quantity` and build the extension
2. Go to the shop archive
3. Add a simple product to the cart multiple times
4. Refresh the page
5. Add the same product to the cart and confirm that the event sent has the correct quantity of `1`
6. Go to the single product page
7. Set the quantity `4` and add it to cart
8. Confirm the quantity sent in the event is `4`

### Changelog entry

> Fix - Incorrect quantity value when adding product to the cart that exists in the cart

### DIFF
--- a/tests/e2e/specs/gtag-events/classic-pages.test.js
+++ b/tests/e2e/specs/gtag-events/classic-pages.test.js
@@ -81,7 +81,7 @@ test.describe( 'GTag events on classic pages', () => {
 
 		await page.goto( `?p=${ simpleProductID }` );
 
-		await page.locator( '.quantity input.qty' ).first().fill('3');
+		await page.locator( '.quantity input.qty' ).first().fill( '3' );
 
 		const addToCart = `.single_add_to_cart_button[value="${ simpleProductID }"]`;
 		const addToCartButton = await page.locator( addToCart ).first();
--- a/tests/e2e/specs/gtag-events/classic-pages.test.js
+++ b/tests/e2e/specs/gtag-events/classic-pages.test.js
@@ -74,6 +74,32 @@ test.describe( 'GTag events on classic pages', () => {
 		} );
 	} );
 
+	test( 'Add to cart quantity is sent on a single product page', async ( {
+		page,
+	} ) => {
+		const event = trackGtagEvent( page, 'add_to_cart' );
+
+		await page.goto( `?p=${ simpleProductID }` );
+
+		await page.locator( '.quantity input.qty' ).first().fill('3');
+
+		const addToCart = `.single_add_to_cart_button[value="${ simpleProductID }"]`;
+		const addToCartButton = await page.locator( addToCart ).first();
+
+		await addToCartButton.click();
+
+		await event.then( ( request ) => {
+			const data = getEventData( request, 'add_to_cart' );
+			expect( data.product1 ).toEqual( {
+				id: simpleProductID.toString(),
+				nm: 'Simple product',
+				ca: 'Uncategorized',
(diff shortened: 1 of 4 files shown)


tests/e2e/specs/gtag-events/classic-pages.test.js
```diff
@@ -74,6 +74,32 @@ test.describe( 'GTag events on classic pages', () => {
...
+		const event = trackGtagEvent( page, 'add_to_cart' );
+
+		await page.goto( `?p=${ simpleProductID }` );
+
+		await page.locator( '.quantity input.qty' ).first().fill( '3' );
+
+		const addToCart = `.single_add_to_cart_button[value="${ simpleProductID }"]`;
+		const addToCartButton = await page.locator( addToCart ).first();
+
+		await addToCartButton.click();
```
ianlin, line 89: ❔ What do you think adding a `page.reload()` for the classic shop as well so it would match the testing instruction in the PR description?

ianlin (approved):
Thanks for fixing it @martynmjones. I tested both block and classic shops and it worked as expected. I'm approving this but just left a comment regarding the E2E test.

tests/e2e/specs/gtag-events/classic-pages.test.js
martynmjones: Thanks for the review @ianlin 🙌 

The error with the product quantity on the single page wasn't triggered with a page reload in the same way the other one was. When the `added_to_cart` hook runs the quantity wasn't being added to the product object so for the test we just need to click the `Add to cart` button and then confirm the quantity is now present in the event. Apologies if that wasn't clear from the testing instructions.

PR #425, opened 2024-05-16
Add integration with WP Consent API

### LINKED ISSUE #424: Integrate Google Analytics for WooCommerce and WP Consent API
## User story
There are a wide range of WordPress plugins that allow merchants to prompt a user to provide tracking consent. Merchants need the ability to easily integrate Google Analytics for WooCommerce and third-party plugins to adhere to Google's Consent Mode requirements.

### Is your feature request related to a problem? 
There is currently no simple way to integrate third-party consent banners with Google Analytics for WooCommerce.

### Describe the solution you'd like 
Integration with WP Consent API as a way of supporting different plugins.

martynmjones:

Closes #424

Adds integration with the [WP Consent API](https://wordpress.org/plugins/wp-consent-api/) plugin to better support different consent banner configurations.

If the plugin is installed on a website then the existing consent state will be sent as an update when [tracking is initialised](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/4c3a0362f33b32f755884b0c03c338f7a1f71b1a/assets/js/src/index.js#L26). For on-page updates, we're [adding an event listener](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/4c3a0362f33b32f755884b0c03c338f7a1f71b1a/assets/js/src/index.js#L60) for `wp_listen_for_consent_change` which is dispatched by WP Consent API when any changes are made.

#### General test instruction

1. Build extension from `add/424-integration-with-wp-consent-api`
2. Install [Complianz](https://wordpress.org/plugins/complianz-gdpr/) and setup a basic consent banner
3. Debug the site using [Tag Assistant](https://tagassistant.google.com/) while logged out
4. Grant consent via consent banner
5. Confirm a `Consent` event is sent with the updated state
6. Trigger events on page i.e visit the shop archive, visit a single product, add a product to cart
7. Confirm the events are all recorded in Tag Assistant and that the consent `On-page Update` is `Granted` (_Under the Consent tab_)
    Screenshot 2024-05-20 at 17 50 37
8. Revoke consent via the consent banner
9. Confirm that the consent state is no longer `Granted` in Tag Assistant

#### Event order testing

By default, the WP Consent API plugin JavaScript loads in the footer meaning we are unable to send the consent update event with the current state until after the page loads.

We need to support calls to `gtag` from on the page for Google Listings & Ads and other third-party extensions. To do that we're including the `wait_for_update` parameter as [recommended by Google](https://developers.google.com/tag-platform/security/guides/consent?consentmode=advanced#async) so that events sent before the consent update will still be recorded. 57ec41e

1. Trigger an event on the page
```php
add_action(
	'wp_head',
	function() {
		echo '<script>gtag( "event", "example_event" );</script>';
	}
);
```
2. Visit the site while debugging via Tag Assistant with consent `Granted`
    1. Confirm the event is sent before the consent update
    2. View the `Consent` tab and confirm that the `On-page Update` is listed as `Granted`

**Without `wait_for_update`:**

Screenshot 2024-05-20 at 15 23 56

**With `wait_for_update`:**

Screenshot 2024-05-20 at 15 26 37

### Additional details:

- Another option was to add the existing consent state to the `ga4w` data object so that immediately after we set the consent defaults the update can be sent. While this would be preferable, caching and minification plugins would cause problems with it which is why we're relying on a purely JS implementation instead
- **Tests will be added in a follow-up PR**

### Changelog entry

> Add - Integration with the WP Consent API plugin

### DIFF
--- a/assets/js/src/index.js
+++ b/assets/js/src/index.js
@@ -3,7 +3,7 @@ import { classicTracking } from './integrations/classic';
 import { blocksTracking } from './integrations/blocks';
 import {
 	setCurrentConsentState,
-	addConsentStateChangeEventListener
+	addConsentStateChangeEventListener,
 } from './integrations/wp-consent-api';
 
 // Wait for 'ga4w:ready' event if `window.ga4w` is not there yet.
--- a/assets/js/src/index.js
+++ b/assets/js/src/index.js
@@ -1,7 +1,10 @@
 import { setupEventHandlers } from './tracker';
 import { classicTracking } from './integrations/classic';
 import { blocksTracking } from './integrations/blocks';
-import { setCurrentConsentState, addConsentStateChangeEventListener } from './integrations/wp-consent-api';
+import {
+	setCurrentConsentState,
+	addConsentStateChangeEventListener
+} from './integrations/wp-consent-api';
 
 // Wait for 'ga4w:ready' event if `window.ga4w` is not there yet.
 if ( window.ga4w ) {
@@ -16,11 +19,6 @@ if ( window.ga4w ) {
 		window.addEventListener( 'load', warnIfDataMissing );
 	}
 }
-
-const consentMap = {
-	statistics: [ 'analytics_storage' ],
-	marketing: [ 'ad_storage', 'ad_user_data', 'ad_personalization' ],
-};
 
 function initializeTracking() {
 	setCurrentConsentState( window.ga4w.settings );
--- a/assets/js/src/index.js
+++ b/assets/js/src/index.js
@@ -15,7 +15,37 @@ if ( window.ga4w ) {
 		window.addEventListener( 'load', warnIfDataMissing );
 	}
 }
+
+const consentMap = {
+	statistics: [
+		'analytics_storage',
+	],
+	marketing: [
+		'ad_storage',
+		'ad_user_data',
+		'ad_personalization',
+	],
+};
+
 function initializeTracking() {
+	if ( typeof wp_has_consent === 'function' ) {
+		window.wp_consent_type = 'optin';
+
+		const consentState = {};
+
+		for ( const [ category, types ] of Object.entries( consentMap ) ) {
+			if ( wp_has_consent( category ) ) {
+				types.forEach( type => {
+					consentState[ type ] = 'granted';
+				} );
+			}
+		}
+
+		if ( Object.keys( consentState ).length > 0 ) {
+			gtag( 'consent', 'update', consentState );
+		}
+	}
+
 	const getEventHandler = setupEventHandlers( window.ga4w.settings );
 
 	classicTracking( getEventHandler, window.ga4w.data );
--- a/assets/js/src/index.js
+++ b/assets/js/src/index.js
@@ -17,32 +17,26 @@ if ( window.ga4w ) {
 }
 
 const consentMap = {
-	statistics: [
-		'analytics_storage',
-	],
-	marketing: [
-		'ad_storage',
-		'ad_user_data',
-		'ad_personalization',
-	],
+	statistics: [ 'analytics_storage' ],
+	marketing: [ 'ad_storage', 'ad_user_data', 'ad_personalization' ],
 };
 
 function initializeTracking() {
-	if ( typeof wp_has_consent === 'function' ) {
+	if ( typeof wp_has_consent === 'function' ) { // eslint-disable-line camelcase
 		window.wp_consent_type = 'optin';
 
 		const consentState = {};
 
 		for ( const [ category, types ] of Object.entries( consentMap ) ) {
-			if ( wp_has_consent( category ) ) {
-				types.forEach( type => {
+			if ( wp_has_consent( category ) ) { // eslint-disable-line camelcase, no-undef
+				types.forEach( ( type ) => {
 					consentState[ type ] = 'granted';
 				} );
 			}
 		}
 
 		if ( Object.keys( consentState ).length > 0 ) {
-			gtag( 'consent', 'update', consentState );
+			gtag( 'consent', 'update', consentState ); // eslint-disable-line no-undef
 		}
 	}
 
@@ -64,16 +58,17 @@ function warnIfDataMissing() {
 document.addEventListener( 'wp_listen_for_consent_change', function ( event ) {
 	const consentUpdate = {};
 
-	const types = consentMap[ Object.keys( event.detail )[0] ];
-	const state = Object.values( event.detail )[0] === 'allow' ? 'granted' : 'deny';
+	const types = consentMap[ Object.keys( event.detail )[ 0 ] ];
+	const state =
+		Object.values( event.detail )[ 0 ] === 'allow' ? 'granted' : 'deny';
 
 	if ( types !== undefined ) {
-		types.forEach( type => {
+		types.forEach( ( type ) => {
 			consentUpdate[ type ] = state;
 		} );
 
 		if ( Object.keys( consentUpdate ).length > 0 ) {
-			gtag( 'consent', 'update', consentUpdate );
+			gtag( 'consent', 'update', consentUpdate ); // eslint-disable-line no-undef
 		}
 	}
-});
+} );
--- a/assets/js/src/index.js
+++ b/assets/js/src/index.js
@@ -1,6 +1,7 @@
 import { setupEventHandlers } from './tracker';
 import { classicTracking } from './integrations/classic';
 import { blocksTracking } from './integrations/blocks';
+import { setCurrentConsentState, addConsentStateChangeEventListener } from './integrations/wp-consent-api';
 
 // Wait for 'ga4w:ready' event if `window.ga4w` is not there yet.
 if ( window.ga4w ) {
@@ -22,25 +23,8 @@ const consentMap = {
 };
 
 function initializeTracking() {
-	// eslint-disable-next-line camelcase
-	if ( typeof wp_has_consent === 'function' ) {
-		window.wp_consent_type = 'optin';
-
-		const consentState = {};
-
-		for ( const [ category, types ] of Object.entries( consentMap ) ) {
-			// eslint-disable-next-line camelcase, no-undef
-			if ( wp_has_consent( category ) ) {
-				types.forEach( ( type ) => {
-					consentState[ type ] = 'granted';
-				} );
-			}
-		}
-
-		if ( Object.keys( consentState ).length > 0 ) {
-			gtag( 'consent', 'update', consentState ); // eslint-disable-line no-undef
-		}
-	}
+	setCurrentConsentState( window.ga4w.settings );
+	addConsentStateChangeEventListener( window.ga4w.settings );
 
 	const getEventHandler = setupEventHandlers( window.ga4w.settings );
 
@@ -56,21 +40,3 @@ function warnIfDataMissing() {
 		);
 	}
 }
-
-document.addEventListener( 'wp_listen_for_consent_change', function ( event ) {
-	const consentUpdate = {};
-
-	const types = consentMap[ Object.keys( event.detail )[ 0 ] ];
-	const state =
-		Object.values( event.detail )[ 0 ] === 'allow' ? 'granted' : 'deny';
-
-	if ( types !== undefined ) {
-		types.forEach( ( type ) => {
-			consentUpdate[ type ] = state;
-		} );
-
-		if ( Object.keys( consentUpdate ).length > 0 ) {
-			gtag( 'consent', 'update', consentUpdate ); // eslint-disable-line no-undef
-		}
-	}
-} );
--- a/assets/js/src/index.js
+++ b/assets/js/src/index.js
@@ -22,13 +22,15 @@ const consentMap = {
 };
 
 function initializeTracking() {
-	if ( typeof wp_has_consent === 'function' ) { // eslint-disable-line camelcase
+	// eslint-disable-next-line camelcase
+	if ( typeof wp_has_consent === 'function' ) {
 		window.wp_consent_type = 'optin';
 
 		const consentState = {};
 
 		for ( const [ category, types ] of Object.entries( consentMap ) ) {
-			if ( wp_has_consent( category ) ) { // eslint-disable-line camelcase, no-undef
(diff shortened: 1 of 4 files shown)


assets/js/src/index.js
```diff
@@ -15,7 +15,33 @@ if ( window.ga4w ) {
...
+			if ( wp_has_consent( category ) ) {
+				types.forEach( ( type ) => {
+					consentState[ type ] = 'granted';
+				} );
+			}
+		}
+
+		if ( Object.keys( consentState ).length > 0 ) {
+			gtag( 'consent', 'update', consentState ); // eslint-disable-line no-undef
+		}
```
tomalec: :construction: I think we should also support scenarios in which the default consent state is other than "deny all."
For example, when a merchant uses the snippet to set other consent modes, we should send a full map update for "granted" and "denied" as well.

assets/js/src/index.js
```diff
@@ -15,7 +15,33 @@ if ( window.ga4w ) {
...
 	}
 }
+
+const consentMap = {
+	statistics: [ 'analytics_storage' ],
+	marketing: [ 'ad_storage', 'ad_user_data', 'ad_personalization' ],
+};
+
 function initializeTracking() {
+	// eslint-disable-next-line camelcase
```
tomalec: :nail_care: 
WDYT of explaining why we disable this rule:
```suggestion
	// eslint-disable-next-line camelcase -- `wp_has_consent` is defined by the WP Consent API plugin.
```

assets/js/src/index.js
```diff
@@ -15,7 +15,33 @@ if ( window.ga4w ) {
...
+	if ( typeof wp_has_consent === 'function' ) {
+		window.wp_consent_type = 'optin';
+
+		const consentState = {};
+
+		for ( const [ category, types ] of Object.entries( consentMap ) ) {
+			// eslint-disable-next-line camelcase, no-undef
```
tomalec: :nail_care: 
```suggestion
			// eslint-disable-next-line camelcase, no-undef -- We checked for the presence of the WP Consent API's function above.
```

assets/js/src/index.js
```diff
@@ -15,7 +15,33 @@ if ( window.ga4w ) {
...
+			if ( wp_has_consent( category ) ) {
+				types.forEach( ( type ) => {
+					consentState[ type ] = 'granted';
+				} );
+			}
+		}
+
+		if ( Object.keys( consentState ).length > 0 ) {
+			gtag( 'consent', 'update', consentState ); // eslint-disable-line no-undef
```
tomalec: :construction: 
This may not be called `gtag` as per https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/4c3a0362f33b32f755884b0c03c338f7a1f71b1a/includes/class-wc-google-gtag-js.php#L94:L94

assets/js/src/index.js
```diff
@@ -15,7 +15,33 @@ if ( window.ga4w ) {
...
 }
+
+const consentMap = {
+	statistics: [ 'analytics_storage' ],
+	marketing: [ 'ad_storage', 'ad_user_data', 'ad_personalization' ],
+};
+
 function initializeTracking() {
+	// eslint-disable-next-line camelcase
+	if ( typeof wp_has_consent === 'function' ) {
```
tomalec: :nail_care: WDYT of moving this entire block into `./integrations/wp-consent-api` and calling it as 

```js
import { updateConsent, listenToConsentChanges } from './integrations/wp-consent-api';
// ...
function initializeTracking() {
   wpConsentAPI( window.ga4w.settings )
   // ...
}

listenToConsentChanges();

```

assets/js/src/index.js
```diff
@@ -30,3 +56,21 @@ function warnIfDataMissing() {
...
+	const state =
+		Object.values( event.detail )[ 0 ] === 'allow' ? 'granted' : 'deny';
+
+	if ( types !== undefined ) {
+		types.forEach( ( type ) => {
+			consentUpdate[ type ] = state;
+		} );
+
+		if ( Object.keys( consentUpdate ).length > 0 ) {
+			gtag( 'consent', 'update', consentUpdate ); // eslint-disable-line no-undef
```
tomalec: :construction: 
The `gtag` may not be set . It could be called differently, as per https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/4c3a0362f33b32f755884b0c03c338f7a1f71b1a/includes/class-wc-google-gtag-js.php#L94:L94

```suggestion
			const gtag = window[ ga4w.settings.tracker_function_name ]; // eslint-disable-line camelcase -- This is a PHP-set config.
			gtag( 'consent', 'update', consentState );
```

includes/class-wc-google-gtag-js.php
```diff
@@ -94,7 +94,7 @@ private function register_scripts(): void {
 					function %2$s(){dataLayer.push(arguments);}
 					// Set up default consent state.
 					for ( const mode of %4$s || [] ) {
-						%2$s( "consent", "default", mode );
+						%2$s( "consent", "default", { ...mode, "wait_for_update": 500 } );
```
tomalec: :question: 
How about letting merchants customize their own `wait_for_update` value?
```suggestion
						%2$s( "consent", "default", { "wait_for_update": 500, ...mode } );
```

tomalec (changes requested):
I run tests locally - works well. However, initially, I get some quirk that WP Consent API's cookie was not created after allowing in Complianz modal. Restarting new incognito mode helped. I'll take some time to try to reproduce it, not to miss something.

I left some comments regarding the code. The most important are marked with :construction:

assets/js/src/index.js
martynmjones: Thanks for catching that. Updated in 4360cce

assets/js/src/index.js
martynmjones: Added in 10c2af7

assets/js/src/index.js
martynmjones: Moved in 4360cce to keep things cleaner.

includes/class-wc-google-gtag-js.php
martynmjones: Nice suggestion! 4b9cc2c

assets/js/src/index.js
martynmjones: 🤔  I'm not convinced this is a scenario we want to cover and if we were to then it would introduce some complexity.

If WP Consent API is installed on a site then by default `wp_has_consent` will return `false` for all categories when `wp_consent_type` is set to `optin`.

So that in itself introduces a few problems. If the consent defaults have been filtered to `granted` then we would send those defaults, check the user's consent state, and then immediately send an update denying all categories if the user hasn't granted consent yet which negates the filter.

Additionally, for regions where we're not setting the default consent state we would still send an update with all categories denied. So if a merchant is selling in say the UK and USA then we would send a consent update for USA users denying all categories where there may not be a consent banner.

We might be able to work around those issues by also checking to see if the consent state cookies for each category is set before we trigger an update but I don't think that's a good approach and we'd be muddying the integration to support an edge case that goes against Google's guidelines.

Our goal is to add support for WP Consent API so that merchants based in regions where consent mode is now required can easily set up a consent banner. In those regions, Google requires explicit consent from the user so I'd argue it's safe to proceed with the assumption that consent defaults are set to deny.

martynmjones:
Hey @tomalec, thanks for the review 🙌 

I've made some adjustments and responded to your comments so this is ready for a second round.

assets/js/src/index.js
tomalec: > If the consent defaults have been filtered to granted then we would send those defaults, check the user's consent state, and then immediately send an update denying all categories if the user hasn't granted consent yet which negates the filter.

Yes. And in my understanding that could be an expected result. The merchant uses our filter to set all as `granted,` then activates or not the WP Consent API to update the state according to preferences set in the modal. Why would that be a problem?

> if a merchant is selling in say the UK and USA then we would send a consent update for USA users denying all categories where there may not be a consent banner.

But currently, we do not deny for USA visitors if there is a modal, and they explicitly set "deny" using that modal.
That, looks like a bug to me.


Steps to reproduce:
1. Use VPN to connect from US 
2. Remove any snippets to check the plain default behavior.
3. Open tagassistant
4. Load the page
5. :green_circle: no default consent state is set for US region
6. On the page, open the consent modal, and click "deny"
8. :red_circle: no consent update is tracked by TA
7. Click "Add to cart"
8. :red_circle: `add_to_cart` event is sent with all empty fields.


The JS is run not for the "site config" but for the particular user, so if they grant or deny, we should acknowledge their choice regardless of their region.

assets/js/src/integrations/wp-consent-api.js
```diff
@@ -0,0 +1,56 @@
...
+
+export const addConsentStateChangeEventListener = ( {
+	tracker_function_name: trackerFunctionName,
+} ) => {
+	document.addEventListener( 'wp_listen_for_consent_change', ( event ) => {
+		const consentUpdate = {};
+
+		const types = consentMap[ Object.keys( event.detail )[ 0 ] ];
+		const state =
+			Object.values( event.detail )[ 0 ] === 'allow' ? 'granted' : 'deny';
```
tomalec: :construction:  It's `denied` not `deny` 
https://developers.google.com/tag-platform/gtagjs/reference

```suggestion
		const state =
			Object.values( event.detail )[ 0 ] === 'allow'
				? 'granted'
				: 'denied';
```

assets/js/src/index.js
tomalec: Ah, I found the reason for the bug above https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/425/files#r1610578378

tomalec: Steps to reproduce the case related to `setCurrentConsentState`:

1. Use VPN to connect from US
2. Remove any snippets to check the plain default behavior.
3. Open tagassistant
4. Load the page
5. 🟢 no default consent state is set for US region
6. On the page, open the consent modal, and click "deny"
7. **Reload the page**
8. 🔴 no consent update is tracked by TA
9. Click "Add to cart"
10. 🔴 add_to_cart event is sent with all empty fields.

assets/js/src/integrations/wp-consent-api.js
```diff
@@ -0,0 +1,56 @@
...
+
+		const consentState = {};
+
+		for ( const [ category, types ] of Object.entries( consentMap ) ) {
+			// eslint-disable-next-line camelcase, no-undef
+			if ( wp_has_consent( category ) ) {
+				types.forEach( ( type ) => {
+					consentState[ type ] = 'granted';
+				} );
+			}
```
tomalec: :nail_care: We can add an explanation to this `eslint-disable-next-line` as well.

:question: Related to https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/425#discussion_r1608711163

```suggestion
			types.forEach( ( type ) => {
				// eslint-disable-next-line camelcase, no-undef
				consentState[ type ] = wp_has_consent( category )
					? 'granted'
					: 'denied';
			} );
```

AFAIUnderstood the docs `gtag('consent', 'update', ... )` is for the current visitor, so the region does not matter. If they disallow, I assume they mean to disallow tracking their data regardless where they connect from at the moment.

tomalec:
After adding the fixes I suggested, I tested the branch with the following extensions:
   1. Complianz, w/ & w/o Autooptimize
   2. Cookie bot, w/ & w/o Autooptimize
   3. GDPR Cookie Compliance, w/ & w/o Autooptimize


I added the WP Consent API steps to general testing wiki https://github.com/woocommerce/woocommerce-google-analytics-integration/wiki/General-Testing#12-wp-consent-api-integration

I also created Setup & Smoke testing instructions for those three extensions in a separate Wiki: https://github.com/woocommerce/woocommerce-google-analytics-integration/wiki/Smoke-Testing-WP-Consent-API-integration-with-other-extensions. I know we don't have the capacity to maintain support for all of them. Still, IMHO, it would be good to at least one integration works.

I made a PR with a small README note about WP Consent API and UI for banners.
https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/426 

-----

> Tests will be added in a follow-up PR

Do you have anything started, or can I pick that up?

assets/js/src/integrations/wp-consent-api.js
martynmjones: Thanks for catching that! Updated in ff8ef0a

assets/js/src/integrations/wp-consent-api.js
martynmjones: Apologies if I wasn't clear in my earlier description @tomalec.

If GA4W is installed on a store (_without the default consent state being filtered_) then for visitors from the EEA region, the on-page default will be `denied`.

A user from a different country then visits the store and the on-page default does not apply to them.

With this suggested change, for visitors from outside the EEA region (_that the on-page default doesn't apply to_) we would still send an update denying all categories even if they haven't interacted with a consent banner because `wp_has_consent` will default to `false`.

Here are two screenshots taken while visiting the site from the USA via a VPN with the suggested change without interacting with a consent banner:

**On-page default:**

Screenshot 2024-05-28 at 11 39 10

**On-page update:**

Screenshot 2024-05-28 at 11 39 17

So here no traffic from outside of the EEA region will be tracked and merchants might not display a banner to those users to change the consent state if it's not a local requirement.

I've adjusted the code to send the entire consent state but only if the category cookie is set to avoid sending the update before a user has interacted with a consent banner. 86e43c7

assets/js/src/index.js
martynmjones: See https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/425#discussion_r1617289542

martynmjones:
Hey @tomalec, many thanks for the additional testing and for updating the documentation. I've made some changes based on our discussions so far if you could please take another look.

assets/js/src/integrations/wp-consent-api.js
tomalec: Thanks for the explanations, now I get it and agree :)

tomalec (approved):
Reviewed the code, tested locally. Using Polish IP, and US one, tested with two cookie banner plugins and fiddling manually with API.

LGTM

PR #428, opened 2024-05-30
Avoid tracking data when user is not opted-in

puntope:

When the user is not opted into WooCommerce tracking we should avoid tracking the data for the extension. However, right now we are not doing that and we track the data without checking if the user is opted-in. This PR fixes that.

### Checks:

### Screenshots:

Screenshot 2024-05-30 at 12 24 59

Screenshot 2024-05-30 at 12 23 45

1. Checkout this PR 
2. Run `wp option update woocommerce_allow_tracking "no"` in WP CLI or set `woocommerce_allow_tracking` option as "no" 
3. Run `wp wc tracker snapshot --format=yaml` in WPCLI 
4. See the `wc-google-analytics` not appearing in the tracking
5. Run `wp option update woocommerce_allow_tracking "yes"` in WP CLI or set `woocommerce_allow_tracking` option as "yes" 
6.  Run `wp wc tracker snapshot --format=yaml` in WPCLI
7. . See the `wc-google-analytics` appearing in the tracking

### Changelog entry

> Fix - Avoid tracking data when user is not opted-in

### DIFF
--- a/includes/class-wc-google-analytics.php
+++ b/includes/class-wc-google-analytics.php
@@ -220,7 +220,7 @@ class WC_Google_Analytics extends WC_Integration {
 	 */
 	public function track_settings( $data ) {
 		if ( 'no' === get_option( 'woocommerce_allow_tracking', 'no' ) ) {
-			return [];
+			return $data;
 		}
 
 		$settings                    = $this->settings;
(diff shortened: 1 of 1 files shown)


tomalec (approved):
Reviewed the code, tested locally, LGTM.
Thanks for fixing that!

puntope:
WC _Tracks is already preventing to send the data when user is not Opted in. So this check is not necessary.

PR #429, opened 2024-05-31
Add E2E tests for the WP Consent API integration

martynmjones:

Follow-up to #425

Adds E2E tests for the integration with WP Consent API.

**Scenarios covered:**

1. GA4W sets the global `wp_consent_type` variable to `optin` if it's not currently set
2. Consent `granted` update is sent for `analytics_storage` when the WP Consent API category `statistics` is `allowed`
3. Consent `granted` update is sent for `ad_storage`, `ad_user_data`, and `ad_personalization` when the WP Consent API category `marketing` is `allowed`
4. The current consent state is sent when the page loads if the default state is `denied` and the current state is `granted`
5. The current consent state is sent when the page loads if the default state is `granted` and the current state is `denied`

_Note: Scenario 2 & 3 cover on page updates via the `wp_listen_for_consent_change` event listener. Scenario 4 & 5 are immediate updates when the page loads based on the current event state._

### Screenshots:

Screenshot 2024-05-31 at 18 24 07

1. Checkout `add/wp-consent-api-integration-tests` and build extension
2. Setup [E2E environment and run tests](https://github.com/woocommerce/woocommerce-google-analytics-integration?tab=readme-ov-file#e2e-testing)
3. Confirm all tests pass

### DIFF
--- a/tests/e2e/specs/js-scripts/wp-consent-api.test.js
+++ b/tests/e2e/specs/js-scripts/wp-consent-api.test.js
@@ -91,7 +91,8 @@ test.describe( 'WP Consent API Integration', () => {
 		await page.evaluate( () =>
 			window.wp_set_consent( 'marketing', 'allow' )
 		);
-		await page.goto( 'shop?consent_default=denied' );
+		// Go to a new page to confirm that the consent state is maintained across page loads
+		await page.goto( '/?consent_default=denied' );
 
 		const dataLayer = await page.evaluate( () => window.dataLayer );
 		const consentState = dataLayer.filter( ( i ) => i[ 0 ] === 'consent' );
--- a/tests/e2e/specs/js-scripts/wp-consent-api.test.js
+++ b/tests/e2e/specs/js-scripts/wp-consent-api.test.js
@@ -1,0 +1,156 @@
+/**
+ * External dependencies
+ */
+const { test, expect } = require( '@playwright/test' );
+
+/**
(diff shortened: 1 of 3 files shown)


tests/e2e/specs/js-scripts/wp-consent-api.test.js
```diff
@@ -0,0 +1,156 @@
...
+	} );
+
+	test( 'Consent state is sent as update when page is loaded', async ( {
+		page,
+	} ) => {
+		await page.goto( 'shop?consent_default=denied' );
+		await page.evaluate( () =>
+			window.wp_set_consent( 'marketing', 'allow' )
+		);
+		await page.goto( 'shop?consent_default=denied' );
```
eason9487: It might be useful to add a code comment to explain why it's going to the same page again. Alternatively, goto a different page to more clearly show the difference from the previous test?

eason9487 (approved):
Thanks for adding the tests. I can run all E2E tests successfully. LGTM.

The only comment is not a blocker and can be ignored.

tests/e2e/specs/js-scripts/wp-consent-api.test.js
martynmjones: Thanks for the review and suggestion @eason9487! Suggested change made in d7100f5

tomalec:
I've added two more tests so we check also revoking the consent.

PR #430, opened 2024-06-03
Add E2E tests for denying via WP Consent API

tomalec:

This is a small add-on to https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/429.
So, we test that a visitor can also deny their consent using the WP Consent API.

### Checks:

### Screenshots:

image

1. Checkout `add/wp-consent-api-integration-tests-2` and build extension
2. Setup [E2E environment and run tests](https://github.com/woocommerce/woocommerce-google-analytics-integration?tab=readme-ov-file#e2e-testing)
3. Confirm all tests pass

### DIFF
--- a/tests/e2e/specs/js-scripts/wp-consent-api.test.js
+++ b/tests/e2e/specs/js-scripts/wp-consent-api.test.js
@@ -84,6 +84,66 @@ test.describe( 'WP Consent API Integration', () => {
 		} );
 	} );
 
+	test( 'Consent update denying `analytics_storage` is sent when WP Consent API `statistics` category is `denied`', async ( {
+		page,
+	} ) => {
+		await page.goto( 'shop?consent_default=granted' );
+		await page.evaluate( () =>
+			window.wp_set_consent( 'statistics', 'deny' )
+		);
+
+		const dataLayer = await page.evaluate( () => window.dataLayer );
+		const consentState = dataLayer.filter( ( i ) => i[ 0 ] === 'consent' );
+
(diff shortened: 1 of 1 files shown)


tests/e2e/specs/js-scripts/wp-consent-api.test.js
```diff
@@ -84,6 +84,66 @@ test.describe( 'WP Consent API Integration', () => {
 		} );
 	} );
 
+	test( 'Consent update denying `analytics_storage` is sent when WP Consent API `statistics` category is `denied`', async ( {
+		page,
+	} ) => {
+		await page.goto( 'shop?consent_default=granted' );
```
eason9487: Might the `consent_default` parameter be omitted? This would be a bit more consistent with this test case:

https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/88f9ea30ac687eccc84248fbeb557081a61b0280/tests/e2e/specs/js-scripts/wp-consent-api.test.js#L184-L191

But if the purpose of specifying this parameter is to explicitly express its initial value, then I think that's also good, and perhaps the parameter could be added to the last test case as well.

eason9487 (approved):
Thanks for adding more test cases. I can run all E2E tests successfully. LGTM.

The only comment is not a blocker

tests/e2e/specs/js-scripts/wp-consent-api.test.js
tomalec: Thanks, I removed it for consistency :)

PR #438, opened 2024-06-15
Do not replace another `added_to_cart` vanilla handler if there is one

tomalec:

Do not replace another `added_to_cart` vanilla handler if there is one.

Or jQuery event to Vanilla listener hack for `added_to_cart` may remove any previous listeners if there are any.

For other events: `onupdated_wc_div` and `onremoved_from_cart` we prevent that.

### Checks:

1. Simulate another extension listening to the event
```php
add_action(
	'wp_head',
	function () {
		wp_add_inline_script(
			'woocommerce-google-analytics-integration',
			'document.body.onadded_to_cart = () => { console.log("Other extension listens to added to cart"); };',
			'before',
		);
	},
);
```
3. Go to shop page
4. Click add to cart
5. You should see the console log, and event should be tracked in https://tagassistant.google.com/

### Changelog entry

> Tweak - Do not replace other extensions' `add_to_cart` listeners.

### DIFF
--- a/assets/js/src/integrations/classic.js
+++ b/assets/js/src/integrations/classic.js
@@ -48,17 +48,15 @@ export function classicTracking(
 	 * @param {Object}        fragments - An object containing fragments of the updated cart.
 	 * @param {string}        cartHash  - A string representing the hash of the cart after the update.
 	 * @param {HTMLElement[]} button    - An array of HTML elements representing the add to cart button.
-	 * @param {Array}         rest      - An array of additional arguments passed to the event handler.
 	 */
-	document.body.onadded_to_cart = (
+	document.body.onadded_to_cart = function (
 		e,
 		fragments,
 		cartHash,
-		button,
-		...rest
-	) => {
+		button
+	) {
 		if ( typeof oldAddedToCart === 'function' ) {
-			oldAddedToCart( e, fragments, cartHash, button, ...rest );
+			oldAddedToCart.apply( this, arguments );
 		}
 		// Get product ID from data attribute (archive pages) or value (single product pages).
 		const productID = parseInt(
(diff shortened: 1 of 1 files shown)


assets/js/src/integrations/classic.js
```diff
@@ -40,15 +40,26 @@ export function classicTracking(
...
-	document.body.onadded_to_cart = ( e, fragments, cartHash, button ) => {
+	document.body.onadded_to_cart = (
+		e,
+		fragments,
+		cartHash,
+		button,
+		...rest
+	) => {
+		if ( typeof oldAddedToCart === 'function' ) {
+			oldAddedToCart( e, fragments, cartHash, button, ...rest );
```
eason9487: If the goal is to minimize the impact on other extensions, it's also possible to carry `this` over correctly. For example, I would expect the following function to be called with [the `body` node](https://github.com/woocommerce/woocommerce/blob/8.9.3/plugins/woocommerce/client/legacy/js/frontend/add-to-cart.js#L116) as `this` instead of the `window` object, even though `this` isn't normally used.

```php
add_action(
	'wp_head',
	function () {
		wp_add_inline_script(
			'woocommerce-google-analytics-integration',
			'document.body.onadded_to_cart = function() { console.log( this, ...arguments ); };',
			'before',
		);
	},
);
```

eason9487 (commented):
Thanks for improving the compatibility with other extensions!

The code works well but I would like to know if it could consider carrying the `this` context as well.

tomalec:
Thanks for catching that. :) Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/438/commits/dd9211fa3df8a91209f7765722badaf18e65ca83. Would you mind another round @eason9487

eason9487 (approved):
Thanks for the further adjustments! LGTM.

PR #439, opened 2024-06-16
Add dedicated integration with Product Bundles

### LINKED ISSUE #219: Product Bundles total cost in Google Analytics doesn't reflect bundled items
### Describe the bug:

When a product bundle is passed to Google, the base cost is considered, but not the bundled items. If the "base cost" is set to `0` and the items within the bundle are set to `priced individually` then the result is that the price will be shown as `0` in Google, when it should be the total of items in the bundle. 

### Steps to reproduce:

1. Create a product bundle using [Product Bundles](https://woocommerce.com/products/product-bundles/?aff=10486&cid=1131038) 
2. Set the cost of the product to 0 [screenshot](https://d.pr/i/5pRT7K)
3. Add bundled items and make sure theyre set to "priced individually" [screenshot](https://d.pr/i/Hd16DS)
4. Check how the product is passed to Google Analytics: 

        gtag('event', 'view_item', {
            'items': [{
                'id': 'bundle-one',
                'name': 'Product Bundle',
                'category': 'Accessories',
                'price': '0',
            }]
        });

### Expected behavior:

The cost of the bundle should reflect the items added to the bundle

### Actual behavior:

The cost of the bundle is universally defined as 0 despite the bundled items 

### Additional details:

Originally reported in 5366001-zen

tomalec:

Add dedicated integration with Product Bundles,
Use `->get_bundle_price( 'min' );` instead of `->get_price()` for bundles.

Fixes https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/219 the simple way.

### Checks:

### Screenshots:

image

1. Install and activate [the Product Bundles (PB) extension](https://woocommerce.com/products/product-bundles/)
2. Add a new bundle product
   image
3. Add bundled products that are priced individually 
   image
4. Publish the product bundle
5. Open the site through https://tagassistant.google.com/
6. Go to the product bundle page
7. Check that `view_item` event is tracked with the correct bundle price
8. Add the bundle to cart
9. Check that `add_to_cart` event is tracked with the correct bundle price
10. Uninstall the PB extension and smoke test for regression.

### Additional details:

1. I didn't add E2E tests, as PB is a paid extension, and public contributors to this repo would not be able to use it to run E2Es locally. We can try mocking PB interface, but then it's not really E2E. I'm open to better ideas :) 

### Changelog entry

> Add - WooCommerce Product Bundles integration.

### DIFF
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -230,6 +230,13 @@ abstract class WC_Abstract_Google_Analytics_JS {
 			}
 		}
 
+		// Integration with Product Bundles.
+		// Get the minimum price, as `get_price` may return 0 if the product is a bundle and the price is potentially a range.
+		// Even a range containing a single value.
+		if ( class_exists( '\WC_Product_Bundle' ) && $product instanceof WC_Product_Bundle && is_callable( [ $product, 'get_bundle_price' ] ) ) {
+			$price = $product->get_bundle_price( 'min' );
+		}
+
 		$formatted = array(
 			'id'         => $product_id,
 			'name'       => $product->get_title(),
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -233,7 +233,7 @@ abstract class WC_Abstract_Google_Analytics_JS {
 		// Integration with Product Bundles.
 		// Get the minimum price, as `get_price` may return 0 if the product is a bundle and the price is potentially a range.
 		// Even a range containing a single value.
-		if ( class_exists( '\WC_Product_Bundle' ) && $product instanceof WC_Product_Bundle && is_callable( [ $product, 'get_bundle_price' ] ) ) {
+		if ( $product->is_type( 'bundle' ) && is_callable( [ $product, 'get_bundle_price' ] ) ) {
 			$price = $product->get_bundle_price( 'min' );
 		}
 
--- a/assets/js/src/utils/index.js
+++ b/assets/js/src/utils/index.js
@@ -21,7 +21,8 @@ export const getProductFieldObject = ( product, quantity ) => {
 		...getProductCategories( product ),
 		quantity: product.quantity ?? quantity,
 		price: formatPrice(
-			product.prices.price,
+			// Use line total for bundled products, if available.
+			product.totals?.line_total || product.prices.price,
 			product.prices.currency_minor_unit
 		),
 		...variantData,


includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -230,6 +230,13 @@ public function get_formatted_product( WC_Product $product, $variation_id = 0, $
 			}
 		}
 
+		// Integration with Product Bundles.
+		// Get the minimum price, as `get_price` may return 0 if the product is a bundle and the price is potentially a range.
+		// Even a range containing a single value.
+		if ( class_exists( '\WC_Product_Bundle' ) && $product instanceof WC_Product_Bundle && is_callable( [ $product, 'get_bundle_price' ] ) ) {
```
martynmjones: Had you considered simplifying this to:

```suggestion
		if ( $product->is_type( 'bundle' ) && is_callable( [ $product, 'get_bundle_price' ] ) ) {
```

Granted another extension may be registering it's own product type called `bundle` but in that scenario the class name may also match and as we have the second check to confirm the `get_bundle_price` method exists I think it would be suitable.

martynmjones (commented):
Hey @tomalec, thanks for working on the fix for bundle pricing!

Events tracked via the classic integration all look to be using the correct values for bundle prices.

There is still an issue with the Block `remove_from_cart` event. When we're [building the product object in JS](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/55ef826168fd8f3f7907b53d15f0e8f106e14c1a/assets/js/src/utils/index.js#L12) we're making use of `product.prices.price` and when a product is removed from cart on a block powered page that value will be the base price:

Screenshot 2024-06-18 at 13 42 53

We could instead rely on `totals.line_total` for the price if it is available which should give us consistent pricing across different integrations.

Screenshot 2024-06-18 at 13 50 08

---

> I didn't add E2E tests, as PB is a paid extension, and public contributors to this repo would not be able to use it to run E2Es locally. We can try mocking PB interface, but then it's not really E2E. I'm open to better ideas :)

I don't have much else to suggest here I'm afraid. We could add a unit test in PHP for the bundled price in `get_formatted_product` which would be a lot easier to mock although that wouldn't cover Blocks.

tomalec:
Thanks, @martynmjones, for catching that!
I applied your suggestions. Speaking of tests, as there are none for `get_formatted_product` so far, WDYT of adding them as a follow-up in the next porter or cooldown cycle?

martynmjones (approved):
Thanks for the changes @tomalec!

Pricing is consistent across different events and integrations now so LGTM ✅ 

> WDYT of adding them as a follow-up in the next porter or cooldown cycle?

Agreed! A follow-up for adding tests sounds good.

tomalec:
Thanks :) Created the follow-up issue to track tests https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/442

PR #443, opened 2024-06-24
Add checks and explicit errors when cart buttons are missing data

tomalec:

Add checks and explicit errors when cart buttons are missing data
to help debugging issues with other extensions manipulating WC Core add & remove from cart buttons or events.

This would help identifying issues like the one reported at https://wordpress.org/support/topic/js-error-on-add-to-cart-when-logged-out/

### Checks:

### Screenshots:

image

1. Build the branch
2. open shop page
3. Run synthetic/malformed cart events
```js
jQuery( document.body ).trigger( 'added_to_cart' );
jQuery( document.body ).trigger( 'removed_from_cart' );
```
8. Check for errors in the console
6. Add an item to the cart using UI.
7. Remove all product ids `document.querySelectorAll('[data-product_id]').forEach((e)=>e.removeAttribute('data-product_id'))`. 8. Remove item from mini cart using UI
8. Check for errors in the console
9. Smoke test not tempered events and buttons for adding to & removing from cart, for regression

### Changelog entry

> Tweak - Add checks and explicit errors when cart buttons are missing data.

### DIFF
--- a/assets/js/src/integrations/classic.js
+++ b/assets/js/src/integrations/classic.js
@@ -63,7 +63,7 @@ export function classicTracking(
 			button?.[ 0 ]?.dataset.product_id || button?.[ 0 ]?.value
 		);
 
-		if ( isNaN( productID ) ) {
+		if ( Number.isNaN( productID ) ) {
 			throw new Error(
 				'Google Analytics for WooCommerce: Could not read product ID from the button given in `added_to_cart` event. Check whether WooCommerce Core events or elements are malformed by other extensions.'
 			);
@@ -104,7 +104,7 @@ export function classicTracking(
 	function removeFromCartHandler( element ) {
 		const productID = parseInt( element.target?.dataset.product_id );
 
-		if ( isNaN( productID ) ) {
+		if ( Number.isNaN( productID ) ) {
 			throw new Error(
 				'Google Analytics for WooCommerce: Could not read product ID from the target element given to remove from cart event. Check whether WooCommerce Core events or elements are malformed by other extensions.'
 			);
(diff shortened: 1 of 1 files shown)


assets/js/src/integrations/classic.js
```diff
@@ -60,14 +60,20 @@ export function classicTracking(
 		}
 		// Get product ID from data attribute (archive pages) or value (single product pages).
 		const productID = parseInt(
-			button[ 0 ].dataset.product_id || button[ 0 ].value
+			button?.[ 0 ]?.dataset.product_id || button?.[ 0 ].value
```
eason9487: If the purpose is to help debugging, perhaps it also needs to consider the case where the `button` parameter is an empty array/jQuery collection.

```js
jQuery( document.body ).trigger( 'added_to_cart', [ , , jQuery('match-nothing') ] );
```

image

assets/js/src/integrations/classic.js
```diff
@@ -60,14 +60,20 @@ export function classicTracking(
...
 		);
 
+		if ( isNaN( productID ) ) {
```
eason9487: Although `parseInt` won't return a type other than number or `NaN`, it would be better to avoid using `isNaN` when possible because it has some [confusing special-case behavior](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description) that leads to semantics ambiguity.

Maybe it could be replaced with `Number.isNaN`.

eason9487 (approved):
Thanks for the improvement. LGTM.

Commented on two minor concerns, and I believe it won't need another round even if there are related changes.

tomalec:
Thanks @eason9487 for the detailed review :)

PR #446, opened 2024-06-28
Retire `phpcs-diff`

eason9487:

This PR retires `phpcs-diff` and add the `phpcs` workflow.

### Checks:

1. `npm run lint:php` to see if the check can pass
2. View [the check result](https://github.com/woocommerce/woocommerce-google-analytics-integration/actions/runs/9710643284/job/26801748611) of the `phpcs` workflow
3. View [the annotations of linting errors](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/446/files/0e2ce9b5ef98549a83d8d1dd581622ddfc790211) added by the `phpcs` workflow
   image


tests/class-unittestsbootstrap.php
```diff
@@ -69,8 +70,10 @@ public function set_path_props() {
 	 * Set server props
 	 */
 	public function set_server_props() {
+		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.MissingUnslash, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
puntope: Do you think we can sanitize the input here instead of ignoring the line?

puntope (approved):
Thanks @eason9487 for opening this PR

I tested it and works

Screenshot 2024-06-28 at 18 16 25


I left one comment regarding the handle of the PHPCS exceptions instead of ignoring them if possible, in this PR or maybe create an Issue for it.

Approved tho since it's not a blocker because it was not introduced in this PR.

tests/class-unittestsbootstrap.php
eason9487: Since this file is used for PHPUnit testing, I thought it would be okay to skip the checks.

I revisited its logic once and found that after the rewrite it didn't have to skip checking. Adjusted in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/446/commits/db107f7d216adbd9798695e9c8bd05ddce288a67.

PR #520, opened 2025-09-26
Skip instead of throwing error when product id is not found for legacy Ajax add-to-cart

### LINKED ISSUE #451: Product id reading issue for variable product add-to-cart(AJAX)
### Describe the bug:
The plugin displays an error when clicking the add-to-cart button for the variable product. This occurs only when using the ajax add-to-cart button.

### Steps to reproduce:
1. Activate plugin and set GA tracking id.
1. Use AJAX for add-to-cart products.
2. Add variable product with multiple variations. 
3. On the single product page, choose a product variation from the dropdown and then click on the add-to-cart button.

### Expected behavior:
The plugin should be able to read the product id. 

### Actual behavior:
The plugin is not able to read product id for variable products.

### Additional details:
Plugin cannot retrieve the product id from the button element for a variable product because the variable product id is set via a hidden input rather than the button value attribute.

Screenshot:
image

triple0t:

Closes #451

This PR skips the rest of the checks and consoles the error message instead of throwing the error, which might interfere with other scripts or processing.

The main aim of this PR is not to change the functionality, but rather to ensure that errors thrown here do not interfere with other scripts that might depend on the same event listener. 

### Checks:

### Changelog entry

> Tweak - Don't throw errors when the product ID is not found for legacy Ajax add-to-cart

### DIFF
--- a/assets/js/src/integrations/classic.js
+++ b/assets/js/src/integrations/classic.js
@@ -60,13 +60,15 @@ export function classicTracking(
 		}
 		// Get product ID from data attribute (archive pages) or value (single product pages).
 		const productID = parseInt(
-			button?.[ 0 ]?.dataset.product_id || button?.[ 0 ]?.value
+			button?.[ 0 ]?.dataset?.product_id || button?.[ 0 ]?.value
 		);
 
 		if ( Number.isNaN( productID ) ) {
-			throw new Error(
+			// eslint-disable-next-line no-console
+			console.error(
 				'Google Analytics for WooCommerce: Could not read product ID from the button given in `added_to_cart` event. Check whether WooCommerce Core events or elements are malformed by other extensions.'
 			);
+			return;
 		}
 
 		// If the current product doesn't match search by ID.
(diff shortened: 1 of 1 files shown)


assets/js/src/integrations/classic.js
```diff
@@ -60,13 +60,15 @@ export function classicTracking(
 		}
 		// Get product ID from data attribute (archive pages) or value (single product pages).
 		const productID = parseInt(
-			button?.[ 0 ]?.dataset.product_id || button?.[ 0 ]?.value
+			button?.[ 0 ]?.dataset?.product_id || button?.[ 0 ]?.value
```
eason9487: Did you find any scenarios where the `dataset` could be nullish? If not, we might not need to use optional chaining (`?.`).

assets/js/src/integrations/classic.js
```diff
@@ -102,12 +104,14 @@ export function classicTracking(
 	 * @param {HTMLElement|Object} element - The HTML element clicked on to trigger this event
 	 */
 	function removeFromCartHandler( element ) {
-		const productID = parseInt( element.target?.dataset.product_id );
+		const productID = parseInt( element.target?.dataset?.product_id );
```
eason9487: Same as https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/520#discussion_r2386803523.

eason9487 (approved):
Thanks for the work! LGTM.

Left two questions regarding the need for using optional chaining operators, but they are not blockers.

assets/js/src/integrations/classic.js
triple0t: > Did you find any scenarios where the dataset could be nullish?

Probably not.
I debated whether to add it or not.

`dataset` being a [DOMStringMap](https://developer.mozilla.org/en-US/docs/Web/API/DOMStringMap) means it would always exist provided browser support is sufficient

However, as we can see from the initial issue, some extensions and theme integrations often fail to follow best practices. 

Hmmmm, 🤔  - Let's remove it.

assets/js/src/integrations/classic.js
triple0t: Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/520/commits/6b1f360db74b4341792a46a1661e7169322919e8

assets/js/src/integrations/classic.js
triple0t: Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/520/commits/6b1f360db74b4341792a46a1661e7169322919e8

PR #529, opened 2025-11-30
Add WooCommerce 10.4+ Interactivity API Mini Cart compatibility

lysyjan:
## Description                                                                                                                                             
                                                                                                                                                             
  WooCommerce 10.4 introduces a new Mini Cart component built with the Interactivity API, replacing the previous React-based implementation. This change breaks two analytics tracking features:
                                                                                                                                                             
  1. **`remove_from_cart` event not firing** - The new Mini Cart doesn't fire the `experimental__woocommerce_blocks-cart-remove-item` action hook            
  2. **`begin_checkout` event with empty cart data** - The checkout hook fires before cart data is loaded from the Store API                                 
                                                                                                                                                             
  ## Changes                                                                                                                                                 
                                                                                                                                                             
  ### Mini Cart removal tracking                                                                                                                             
                                                                                                                                                             
  Implemented a click event listener as a fallback for tracking item removals from the Interactivity API Mini Cart. The implementation uses multiple data sources with graceful fallbacks:
                                                                                                                                                             
  1. Interactivity API context (`data-wp-context` attributes)                                                                                                
  2. WooCommerce Blocks store (`wp.data.select('wc/store/cart')`)                                                                                            
  3. Server-rendered cart data (`window.ga4w.data.cart`)                                                                                                     
  4. DOM extraction (product name, price, quantity from visible elements)                                                                                    
                                                                                                                                                             
  A flag mechanism prevents duplicate events when the original hook does fire (for backwards compatibility with older WooCommerce versions).                 
                                                                                                                                                             
  ### Checkout event fallback                                                                                                                                
                                                                                                                                                             
  Added fallback to server-rendered cart data when the Store API cart data is empty or not yet loaded.                                                       
                                                                                                                                                             
  ## Testing                                                                                                                                                 
                                                                                                                                                             
  Requires WooCommerce 10.4.0 Beta 1 or later.                                                                                                               
                                                                                                                                                             
  1. Add a product to cart from the shop page (without page reload)                                                                                          
  2. Open the Mini Cart                                                                                                                                      
  3. Remove the item → `remove_from_cart` event should fire                                                                                                  
  4. Add item again and proceed to checkout → `begin_checkout` event should include cart items                                                               
                                                                                                                                                             
  ## Note                                                                                                                                                    
                                                                                                                                                             
  This is a workaround for missing hook support in WooCommerce's new Interactivity API Mini Cart. Ideally, WooCommerce should fire the `experimental__woocommerce_blocks-cart-remove-item` action for backwards compatibility.

### DIFF
--- a/assets/js/src/integrations/blocks.js
+++ b/assets/js/src/integrations/blocks.js
@@ -9,6 +9,13 @@ import { ACTION_PREFIX, NAMESPACE } from '../constants';
 let hookFiredRecently = false;
 
 /**
+ * Get WooCommerce settings from the global wcSettings object.
+ *
+ * @return {Object|undefined} The wcSettings object or undefined if not available.
+ */
+const getWcSettings = () => window.wc?.wcSettings || window.wcSettings;
+
+/**
  * Get the current cart data from WooCommerce's store (if available) or fallback to static data.
  * The store provides fresh cart data that updates when items are added/removed via AJAX.
  *
@@ -30,41 +37,52 @@ const getCartData = () => {
 };
 
 /**
- * Extract product data from Interactivity API context on a cart item element.
- * WooCommerce 10.4+ stores cart item data in data-wp-context attributes.
+ * Parse a price string into a numeric value using WooCommerce's accounting.js library.
+ * Falls back to wcSettings currency configuration if accounting.js is not available.
  *
- * @param {Element} cartItem - The cart item row element.
- * @return {Object|null} Product object or null if not found.
+ * @param {string} priceText - The raw price text from DOM (e.g., "$1,234.56" or "1.234,56 €").
+ * @return {number} The parsed price as a float, or 0 if parsing failed.
  */
-const getProductFromInteractivityContext = ( cartItem ) => {
-	// Find the element with data-wp-context (could be on cartItem or a parent)
-	const contextElement =
-		cartItem.closest( '[data-wp-context]' ) ||
-		cartItem.querySelector( '[data-wp-context]' );
-
-	if ( ! contextElement ) {
-		return null;
+const parsePriceFromDOM = ( priceText ) => {
+	if ( ! priceText ) {
+		return 0;
 	}
 
-	try {
-		const context = JSON.parse(
-			contextElement.getAttribute( 'data-wp-context' )
-		);
-		// The context structure may vary, look for cart item data
-		const item = context?.woocommerce?.cart?.item || context?.item;
-		if ( item ) {
-			return item;
-		}
-	} catch ( e ) {
-		// Invalid JSON, ignore
+	const wcSettings = getWcSettings();
+	const decimalSeparator = wcSettings?.currency?.decimalSeparator || '.';
+
+	// Use WooCommerce's accounting.js library if available
+	if ( window.accounting?.unformat ) {
+		return window.accounting.unformat( priceText, decimalSeparator );
 	}
 
-	return null;
+	// Fallback: manual parsing using wcSettings currency configuration
+	const thousandSeparator = wcSettings?.currency?.thousandSeparator || ',';
+
+	// Remove currency symbols and whitespace, keeping only digits and separators
+	let cleaned = priceText.replace( /[^\d.,\s]/g, '' ).trim();
+
+	// Remove thousand separators (need to escape special regex chars like '.')
+	const escapedThousand = thousandSeparator.replace(
+		/[.*+?^${}()|[\]\\]/g,
+		'\\$&'
+	);
+	cleaned = cleaned.replace( new RegExp( escapedThousand, 'g' ), '' );
+
+	// Convert decimal separator to standard period for parseFloat
+	if ( decimalSeparator !== '.' ) {
+		cleaned = cleaned.replace( decimalSeparator, '.' );
+	}
+
+	return parseFloat( cleaned ) || 0;
 };
 
 /**
  * Extract product data from DOM elements in a cart item row.
- * This is a fallback when other data sources are unavailable.
+ * This is a last-resort fallback when:
+ * - WooCommerce Blocks store (wc/store/cart) is unavailable or empty
+ * - Static cart data from server is unavailable
+ * - Product cannot be matched by ID or name in cart data
  *
  * @param {Element} cartItem - The cart item row element.
  * @return {Object|null} Product object or null if extraction failed.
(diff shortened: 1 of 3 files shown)


NeosinneR:
@lysyjan is this ready for review or waiting for more work? The CI is misleading, as the required PHP jobs won't start as there are no PHP changes, so you can ignore that (I'll try to take a look at how to configure this properly sometime this week).

lysyjan:
@NeosinneR I mentioned it in the Porter report, as I am not sure about the implementation at all. It was more like a quick exploration of the possible solution for why acceptance tests are failing with Woo 10.4+

NeosinneR:
@lysyjan does the plugin break in WooCommerce 10.4 without this change? 10.4 is aimed to be released tomorrow, but because the PR is in draft and no one is assigned, I don't see any action needed to be done by anyone. But this is the last week for plugin releases before holiday, so if this is needed, we should review, merge and release it ideally this week. If it's not required, it can wait, but it would be still good to know what are next steps with this PR.

lysyjan:
> does the plugin break in WooCommerce 10.4 without this change? 

I don't think so. Based on the code, it looks like the tracking in the mini cart will stop working.

I haven't found this issue during manual smoke tests. I can take a closer look later today, or you can check it manually.
I have noticed that those e2e tests are not very stable.

assets/js/src/integrations/blocks.js
```diff
@@ -1,7 +1,206 @@
...
+		// Find the cart item container to get product information
+		const cartItem = removeButton.closest( '.wc-block-cart-items__row' );
+		if ( ! cartItem ) {
+			return;
+		}
+
+		let product = null;
+
+		// 1. Try Interactivity API context first (WooCommerce 10.4+)
+		product = getProductFromInteractivityContext( cartItem );
```
eason9487: Could you share what conditions make it possible to get the product via `getProductFromInteractivityContext`? I've tried several different combinations of pages and themes, but haven't found it.

assets/js/src/integrations/blocks.js
```diff
@@ -1,7 +1,206 @@
...
+						( item ) => item.name === productName
+					);
+				}
+			}
+		}
+
+		// 3. Fallback: extract from DOM elements
+		if ( ! product ) {
+			product = getProductFromDOM( cartItem );
+		}
```
eason9487: Similar to the above, I also can't find the conditions that would enter this logic branch. Could you share them?

assets/js/src/integrations/blocks.js
```diff
@@ -1,7 +1,206 @@
...
+
+	// Build a minimal product object that works with getProductFieldObject
+	return {
+		name: productName,
+		quantity,
+		prices: {
+			price: Math.round( price * 10 ** currencyMinorUnit ),
+			currency_minor_unit: currencyMinorUnit,
+		},
+	};
```
eason9487: This price parsing logic will have issues when thousand separators are present. The following table shows that common formats can lead to significant discrepancies from the actual values.

| Actual price | `textContent` | `priceText` | `price` |
| :--- | :--- | :--- | :--- |
| 1234 | '$ 1,234' | '1.234' | 1.234 |
| 1234.56 | '$ 1,234.56' | '1.234.56' | 1.234 |
| 1234567 | '¥ 1,234,567' | '1.234,567' | 1.234 |
| 1234 | '1.234 €' | '1.234' | 1.234 |
| 1234.56 | '1.234,56 €' | '1.234.56' | 1.234 |
| 1234567 | '1.234.567 €' | '1.234.567' | 1.234 |

Furthermore, if a WooCommerce store is configured with a precision higher than two decimal places, the hardcoded `currencyMinorUnit` may cause precision loss during the final calculation `Math.round( price * 10 ** currencyMinorUnit )`.

For example, a value of '4.567' would result in
```js
{ price: 457, currency_minor_unit: 2 }
```
instead of
```js
{ price: 4567, currency_minor_unit: 3 }
```

eason9487 (commented):
Thanks for the work! I would like to clarify some conditions of logical branches.

In addition, the`getProductFromDOM` function may not handle common price formats well.

I haven't further verified whether the data from `JSON.parse( document.querySelector( '#wp-script-module-data-\\@wordpress\\/interactivity' ).innerText )` could replace some of the processing, but if not already considered, it might be a potential cart data source.

assets/js/src/integrations/blocks.js
lysyjan: You're right - this function never works. WooCommerce's Interactivity API Mini Cart doesn't store cart item data in data-wp-context attributes. I've removed it and simplified to just use the cart store data with DOM fallback.

assets/js/src/integrations/blocks.js
lysyjan: The DOM fallback triggers when cart store lookup fails - either because wc/store/cart returns empty items or product matching by ID/name fails. In practice this is rare since the cart store is always available when Mini Cart is present. It's mainly a safety net for edge cases.

lysyjan:
@eason9487 Thank you for your great review.

I added a commit updating parsing prices and added an E2E test. I also removed redundant code and updated comments.

assets/js/src/integrations/blocks.js
```diff
@@ -1,7 +1,228 @@
...
+	);
+	cleaned = cleaned.replace( new RegExp( escapedThousand, 'g' ), '' );
+
+	// Convert decimal separator to standard period for parseFloat
+	if ( decimalSeparator !== '.' ) {
+		cleaned = cleaned.replace( decimalSeparator, '.' );
+	}
+
+	return parseFloat( cleaned ) || 0;
+};
```
eason9487, line 82: Finally find a scenario that falls into this case:
1. Use the Twenty Twenty-Two theme
2. Set currency options with `.` as the thousand separator and `,` as the decimal separator
3. Visit the Shop page with an empty cart
4. Add a product to the cart
5. Remove the product via the mini cart

Since `getWcSettings()` returns `undefined`, it will still be parsed as an incorrect price.

tests/e2e/specs/gtag-events/blocks-pages.test.js
```diff
@@ -125,13 +125,122 @@ test.describe( 'GTag events on block pages', () => {
...
+			.first()
+			.click();
+
+		await event.then( ( request ) => {
+			const data = getEventData( request, 'remove_from_cart' );
+			// DOM fallback should parse European format correctly
+			expect( data.product1.nm ).toEqual( 'Simple product' );
+			expect( data.product1.qt ).toEqual( '1' );
+			// Price should be parsed correctly regardless of format
+			expect( parseFloat( data.product1.pr ) ).toBeGreaterThan( 0 );
```
eason9487: Using `toBeGreaterThan( 0 )` may result in a false negative. After changing to `expect( data.product1.pr ).toEqual( simpleProductPrice.toString() );`, this test case will fail because the `wcSettings.currency` injected into the frontend via `page.evaluate` does not affect the DOM already rendered by PHP.

image

eason9487 (commented):
Thanks for the update!

There are still a few issues to resolve regarding the parsing of the price.

lysyjan:
I didn't know that `wcSettings` can be undefined and it can depend on the theme.
I decided to change the architecture to avoid missing currency settings and added them as a part of the script from the plugin.

eason9487 (approved):
Thanks for the work! Tested several scenarios and all functioned normally.

PR #553, opened 2026-02-26
Add npm scripts to run PHPUnit tests via wp-env

### LINKED ISSUE #552: Add npm script to run PHPUnit tests via wp-env
### Description

Running PHPUnit tests locally currently requires a manual multi-step setup: MySQL server, svn, and running `bin/install-unit-tests.sh` to download WordPress core test files and WooCommerce into `/tmp/`. This is fragile and a pain to maintain across dev machines.

We already have a `wp-env` setup (`.wp-env.json`) for E2E tests that spins up WordPress + WooCommerce in Docker. We should leverage that same environment to run PHPUnit tests too, removing the need for local MySQL, svn, and the install script dance.

### Proposed changes

1. **Add a `phpunit.xml.dist` override or env-aware bootstrap** that works inside the wp-env container (the WordPress test library is available at `/wordpress-phpunit/` inside wp-env containers).

2. **Update `.wp-env.json`** to include any additional configuration needed for PHPUnit (e.g. mapping the `tests/` directory, ensuring `phpunit` is available in the container).

3. **Add npm scripts** such as:
   - `npm run test:php` — run PHPUnit tests inside the wp-env container
   - Possibly `npm run test:php -- --filter=TestName` for running specific tests

4. **Update documentation** in `README.md` and `AGENTS.md` to reflect the simpler workflow.

### Expected outcome

Running PHPUnit tests locally should be as simple as:

```sh
npm run wp-env:up
npm run test:php
```

No local MySQL, svn, or `bin/install-unit-tests.sh` required.

### Context

- The CI workflow (`.github/workflows/php-unit-tests.yml`) uses `install-unit-tests.sh` with a real MySQL service — that flow stays as-is for CI.
- This is about improving the **local dev experience** only.
- `wp-env` already includes WordPress test files and a MySQL database in its containers, so the pieces are all there.
- The bootstrap (`tests/class-unittestsbootstrap.php`) hardcodes paths to `/tmp/wordpress-tests-lib` — it will need to detect the wp-env environment or accept overrides via env vars.

### Changelog entry

> Dev - Add npm script to run PHPUnit tests via wp-env for simpler local development.

nerrad:

Closes #552.

Adds `npm run test:php` and `npm run test:php:setup` scripts so PHPUnit tests can run inside the existing wp-env Docker environment. No local MySQL, svn, or manual `bin/install-unit-tests.sh` setup required.

**What changed:**
- **`tests/class-unittestsbootstrap.php`** — `set_path_props()` detects wp-env via the `WP_TESTS_DIR` env var (already set in wp-env containers) and uses container paths. Falls back to the original `sys_get_temp_dir()` paths when not in wp-env, so CI is unaffected.
- **`tests/bin/setup-phpunit-env.sh`** (new) — Lightweight setup script that ensures WooCommerce is installed, downloads 7 specific WC test framework files from GitHub (not included in the wordpress.org zip), and runs `composer install` for the plugin's dev deps.
- **`package.json`** — Two new scripts: `test:php:setup` and `test:php`.
- **`README.md`** / **`AGENTS.md`** — Updated docs.

### Checks:

1. `npm run wp-env:up`
2. `npm run test:php:setup` — should install WooCommerce, download test files, install composer deps
3. `npm run test:php` — should run all 7 PHPUnit tests and pass
4. Optionally test with a specific WC version: `npm run test:php:setup -- --wc-version=9.6.0`
5. Verify CI still passes (the bootstrap change is backward-compatible)

### Changelog entry

> Dev - Add npm scripts to run PHPUnit tests via wp-env for simpler local development.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,10 +48,12 @@ npm run lint:php      # PHPCS (WordPress Coding Standards)
 ### Test
 
 ```sh
-vendor/bin/phpunit                # Run PHP unit tests (see README.md for setup requirements)
 npm run wp-env:up                 # Start local WordPress Docker environment
+npm run test:php:setup            # Install WooCommerce + deps for PHPUnit (once after wp-env:up)
+npm run test:php                  # Run PHP unit tests via wp-env
 npm run test:e2e                  # Run Playwright E2E tests (headless, requires wp-env)
 npm run test:e2e-dev              # Run E2E tests in debug mode with browser visible
+vendor/bin/phpunit                # Run PHP unit tests locally (see README.md for setup)
 ```
 
 ### Local Environment (wp-env)
--- a/README.md
+++ b/README.md
@@ -24,12 +24,27 @@ The `engines` in package.json includes npm `^9` to allow dependabot to update ou
 -   See https://github.com/dependabot/dependabot-core/issues/9277
 
 ## Unit tests
-### Running PHP unit tests in your local dev environment
+
+### Running PHP unit tests via wp-env (recommended)
+
+This uses the same Docker environment as E2E tests — no local MySQL, svn, or manual setup required.
+
+1. Make sure Docker is running
+2. `npm run wp-env:up` to start the environment
+3. `npm run test:php:setup` to install WooCommerce from source and dependencies (only needed once after starting the environment)
+4. `npm run test:php` to run all PHPUnit tests
+
+To test against a specific WooCommerce version:
+
+`npm run test:php:setup -- --wc-version=9.6.0`
+
+### Running PHP unit tests locally (without Docker)
+
 1. Install prerequisites: composer, git, xdebug, svn, wget or curl, mysqladmin
 2. `cd` into the `woocommerce-google-analytics-integration/` plugin directory
 3. Run `composer install`
 4. Run `bin/install-unit-tests.sh <db-name> <db-user> <db-pass> [db-host] [wp-version] [wc-version] [skip-database-creation]` e.g. `bin/install-unit-tests.sh wordpress_test root root localhost latest latest`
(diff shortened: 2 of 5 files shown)


pavel-mailpoet (approved):
Thank you Darren for adding this commands. Tested all the new commands locally. Everything works as described:

```
npm run wp-env:up           ✅
npm run test:php:setup      ✅  (installs WC, downloads test framework files, composer install)
npm run test:php            ✅  (7 tests, 24 assertions pass)
npm run test:php -- --filter=test_scripts_are_registered  ✅  (argument passthrough works)
npm run test:php:setup -- --wc-version=9.6.0              ✅  (version switch works)
```

A few suggestions for hardening:

1. **Validate `--wc-version` input** — The version string is interpolated into a `bash -c` heredoc via `RAW_BASE`. A single-quote in the version value could break out of the inner quoting. Easy fix with a regex guard:
   ```bash
   if [[ ! "$WC_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
       echo "Error: Invalid version format: ${WC_VERSION}" >&2
       exit 1
   fi
   ```

2. **Use `curl -fSs` instead of `curl -sf`** — Currently, if a download fails (e.g. tag doesn't exist on GitHub), the error is completely silent. Adding `-S` (show errors) would print the HTTP error while still suppressing the progress bar.

3. **Add a post-download check** — After the download loop, verify at least one critical file exists (e.g. `tests/legacy/bootstrap.php`) so the user gets a clear error instead of a cryptic PHPUnit failure later.

4. **Composer output truncation** — `| tail -3` on the composer install hides useful error info if something goes wrong. Maybe `| tail -10` or just show full output.

These are all nice-to-haves — the core functionality works well and is a big improvement over the old MySQL/svn setup. 👍

PR #606, opened 2026-06-04
Move tracking settings to instances

### LINKED ISSUE #412: Make methods that use settings non-static
## ~User~ developer story

As discussed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398#discussion_r1540147719

We use a lot of static functions that are not 100% "static," IMO.
[_PHP spec_](https://www.php.net/manual/en/language.oop5.static.php)
> Declaring class properties or methods as static makes them accessible without needing an instantiation of the class.

- `public static function get_enabled_events` uses `self::get`
- `public static function load_opt_out` uses `self::get`
- `public static function get_product_identifier` uses `self::get`
- `public static function get` uses the `self::settings`
**but** we set the `self::settings` in the instance constructor https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/cdb60c08a9cb62f351f5d324472ad62b6eef7105/includes/class-wc-google-gtag-js.php#L47:L47

So we could change `self::get( * )` to `$this->get()` or even `$this->settings[ * ]` (IMHO, `get` just blurs the picture here w/o much benefit).

### Is your feature request related to a problem? 

It's not a big problem. This is a singleton, so the line between class and instance is blurred anyway. However, the `static` could mislead developers by suggesting no instance needs to be created to use them, which is not true.

#### How to reproduce the problem

Use `WC_Google_Gtag_JS::get` without creating the instance first.

### Describe the solution you'd like 

1. Turn `settings` to be instance propery
2. Turn all methods that use `settings` directly or indirectly to be non-`static`
3. Remove `get` function, as currently it limits the data types to be used as settings to `string` which may be problematic in the future

### Describe alternatives you've considered 

Do only 1-2 above.

## Unknowns

- There may be some limitations, logic, or convention put around `settings` and `get` in the WC Core class `WC_Integration`

NeosinneR:

Closes #412
Closes STORMA-41

Moves tracking settings from static storage to the gtag instance and routes internal settings-dependent logic through instance methods. The old public static methods remain as compatibility wrappers.

### Checks:
* [x] Does your code follow the WordPress coding standards?
* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

### Screenshots:

N/A

1. Run `npm run lint:php`.
2. Run `npm run test:php`.

### Additional details:

N/A

### Changelog entry

> Dev - Store Google Analytics tracking settings on the tracking instance.

### Bugs (2)

1. GTAG loader URL may include an empty id parameter (includes/class-wc-google-gtag-js.php)
   - When ga_id is absent, concatenation into the gtag loader URL produces a URL ending with `?id=` (empty measurement id), which is undesirable.

2. Empty ga_id handling in opt-out/inline JS (includes/class-wc-abstract-google-analytics-js.php & includes/class-wc-google-gtag-js.php)
   - get_setting('ga_id') can be null and is passed through esc_js() (coerced to an empty string), resulting in JS like `var gaProperty = '';` and `ga-disable-` (trailing dash) which may cause incorrect opt-out behavior or confusing globals.

### DIFF
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -18,7 +18,7 @@ abstract class WC_Abstract_Google_Analytics_JS {
 	protected static $instance;
 
 	/** @var array $settings Inherited Analytics settings */
-	protected static $settings;
+	protected $settings = array();
 
 	/** @var string Developer ID */
 	public const DEVELOPER_ID = 'dOGY3NW';
@@ -120,7 +120,7 @@ abstract class WC_Abstract_Google_Analytics_JS {
 		add_action(
 			'woocommerce_thankyou',
 			function ( $order_id ) {
-				if ( 'yes' === self::get( 'ga_ecommerce_tracking_enabled' ) ) {
+				if ( 'yes' === $this->get_setting( 'ga_ecommerce_tracking_enabled' ) ) {
 					$order = wc_get_order( $order_id );
 					if ( $order && $order->get_meta( '_ga_tracked' ) !== '1' ) {
 						// Check order key.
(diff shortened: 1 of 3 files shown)


NeosinneR:
@cursor review

NeosinneR:
@cursor review

PR #611, opened 2026-06-04
Defer tracker bundle loading

### LINKED ISSUE #411: Consider using WP Script Modules API
## User story

As discussed in during the review of https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/398#discussion_r1540078257 our scripts would be more async and defer friendly. This could let us use the latest WP [Script Modules API](https://github.com/WordPress/gutenberg/issues/55942).

This could let us use a bit more modern standard, and write the code closer to the one actually executed in the browser.

Unfortunately, for now, we cannot switch yet.
1. The API will be available in WP 6.5, which is not released yet, and we need to support older versions.
2. [WP API does not support inline modules](https://make.wordpress.org/core/2024/03/04/script-modules-in-6-5/#comment-46399). And we currently end data using inline scripts. So we would either need to change that approach, or wait for the support.
3. [The API does not support classic script dependencies](https://make.wordpress.org/core/2024/03/04/script-modules-in-6-5/#comment-46400), so we need to depend on classic inline script data (or change the approach). Most importantly, we depend on `google-tag-manager,` which needs to stay there.

### Describe alternatives you've considered 

Use classic scripts as we use today.

NeosinneR:
## Description
Closes #411.

WordPress Script Modules are available now, but this bundle still relies on classic WordPress package globals and the existing classic script handle is part of the inline-data compatibility surface. Instead of moving to modules in a risky partial migration, this keeps the classic queue and requests WordPress `defer` loading for the main tracker bundle.

## Checks
- [x] Project linting passes
- [x] Automated tests pass
- [x] Manual testing completed

## Testing instructions
1. `npm run lint:php`
2. `npm run test:php`
3. Seed a product if the E2E test site is empty: `npm run -- wp-env run tests-cli wp wc product create --name="Load order product" --type=simple --regular_price=9.99 --user=admin`
4. `npx playwright test --config=tests/e2e/config/playwright.config.js tests/e2e/specs/js-scripts/load-order.test.js`

## Changelog
Update - Request deferred loading for the main tracker bundle while preserving classic script compatibility.

### DIFF
--- a/includes/class-wc-google-gtag-js.php
+++ b/includes/class-wc-google-gtag-js.php
@@ -118,7 +118,10 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 				'google-tag-manager',
 			),
 			Plugin::get_instance()->get_js_asset_version( 'main' ),
-			true
+			array(
+				'in_footer' => true,
+				'strategy'  => 'defer',
+			)
 		);
 	}
 
@@ -130,8 +133,8 @@ class WC_Google_Gtag_JS extends WC_Abstract_Google_Analytics_JS {
 	 */
 	public function enqueue_tracker(): void {
 		wp_enqueue_script( 'google-tag-manager' );
-		// tracker.js needs to be executed ASAP, the remaining bits for main.js could be deferred,
-		// but to reduce the traffic, we ship it all together.
+		// Keep the tracker bundle in the classic script queue so inline data and
+		// classic WordPress package dependencies remain compatible.
 		wp_enqueue_script( $this->script_handle );
 	}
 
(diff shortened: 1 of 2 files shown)


NeosinneR:
@cursor review

NeosinneR:
Closing this after review — **the requested `defer` strategy is a silent no-op, so this PR ships no behavioral change.**

`main.js` is registered with `strategy => 'defer'`, but the inline-data script (`$data_script_handle` in `inline_script_data()`) declares `main.js` as a dependency **and** attaches a blocking `after` inline script (`window.ga4w = {...}; document.dispatchEvent(new Event("ga4w:ready"))`). Per WP core's `WP_Scripts::get_eligible_loading_strategy()` / `filter_eligible_strategies()`, a script is **not** eligible for `defer` when a dependent carries a blocking `after` inline script — WP downgrades `main.js` back to blocking. The rendered `<script>` tag is therefore byte-for-byte identical to `trunk` (`in_footer => true`, no `defer`).

The accompanying unit assertion only checks the *registered* `strategy` data (`wp_scripts()->get_data( 'main', 'strategy' )`), not the rendered tag, so it passes green while the tag never actually carries `defer` — a false-positive signal.

Genuinely enabling defer would require restructuring the inline-data handshake so no blocking `after` inline sits on a dependent of `main.js`. That reopens the exact event-ordering / missed-event risk the current design deliberately avoids, and would need new E2E coverage — out of scope for the low-risk intent here.

Closing rather than merging a no-op. The underlying #411 (Script Modules migration) can be revisited when a real load-strategy improvement justifies the handshake redesign.

PR #650, opened 2026-08-07
Fix purchase item discount and value tax/quantity basis

### LINKED ISSUE #641: purchase item discounts mix tax-inclusive unit prices with tax-exclusive line totals
### Prerequisites

- The problem is reproducible with Google Analytics for WooCommerce 2.3.0.
- Google Tag Assistant captured exactly one `purchase` hit for each fresh order.
- The orders used local pickup with zero shipping.
- The product is a simple product with a catalog price of SEK 95 including 6% tax.
- No customer data, order keys, payment identifiers, GA client IDs, or unmasked URLs are included below.

### Describe the bug

`purchase` item data compares values that use different tax and quantity bases:

- `get_formatted_product()` supplies `prices.price` from `WC_Product::get_price()`. In a store where prices are entered including tax, this is a tax-inclusive unit catalog price.
- `get_formatted_order()` supplies `price_after_coupon_discount` from `WC_Order_Item_Product::get_total()`. This is the complete tax-exclusive line total after coupon discounts.
- `getProductFieldObject()` compares the two values directly. If the order-line total is lower, the difference is emitted as the GA4 item `discount` and the line total is emitted as the item `price`.

This mixes:

1. tax-inclusive and tax-exclusive values; and
2. a unit price and a complete line total.

The same purchase builder also emits `order.totals.total_price` as event-level `value`. That includes tax and may include shipping, while the GA4 event reference defines `value` as the sum of `price × quantity` for the items, excluding tax and shipping.

### Steps to reproduce

Configure WooCommerce to enter catalog prices including tax. Use a simple product priced at SEK 95 including 6% tax.

#### Case 1: quantity 1, no coupon

1. Purchase one unit without a coupon and with zero shipping.
2. Inspect the `purchase` hit in Tag Assistant.

Checkout:

- Gross total: 95
- Tax: 5.38
- Shipping: 0

Observed item and event values:

```text
quantity: 1
price: 89.62
discount: 5.38
value: 95
tax: 5.38
shipping: 0
```

The reported discount equals the tax even though no discount exists.

#### Case 2: quantity 2, no coupon

1. Purchase two units without a coupon and with zero shipping.
2. Inspect the `purchase` hit.

Checkout:

- Gross total: 190
- Tax: 10.75
- Shipping: 0

Observed item and event values:

```text
quantity: 2
price: 95
discount: omitted
value: 190
tax: 10.75
shipping: 0
```

The false discount disappears only because the tax-exclusive line total is now greater than the tax-inclusive unit catalog price. The emitted item price is consequently tax-inclusive.

#### Case 3: quantity 1, 10% coupon

1. Apply a 10% coupon to one unit.
2. Purchase with zero shipping.
3. Inspect the `purchase` hit.

Checkout:

- Gross discount: 9.50
- Gross total: 85.50
- Tax: 4.84
- Shipping: 0

Observed item and event values:

```text
quantity: 1
price: 80.66
discount: 14.34
value: 85.50
tax: 4.84
shipping: 0
```

The correct tax-exclusive unit discount is 8.96. The emitted 14.34 consists of the real tax-exclusive discount, 8.96, plus the original tax, 5.38.

No event-level or item-level `coupon` parameter was emitted. That is not required to reproduce the price and discount defects.

### Expected behavior

All compared item values should use the same tax and quantity basis:

- derive the pre-coupon unit amount from the order line subtotal divided by quantity;
- derive the post-coupon unit amount from the order line total divided by quantity;
- keep both amounts tax-exclusive;
- emit `price` as the post-discount unit price;
- emit `discount` as the pre-discount unit price minus the post-discount unit price;
- omit `discount` when that difference is zero;
- calculate event-level `value` from the emitted item prices multiplied by their quantities, excluding tax and shipping.

Expected values for case 1:

```text
quantity: 1
price: 89.62
discount: omitted or 0
value: 89.62
tax: 5.38
shipping: 0
```

Expected values for case 3:

```text
quantity: 1
price: 80.66
discount: 8.96
value: 80.66
tax: 4.84
shipping: 0
```

For quantities where a tax-exclusive line total cannot be divided evenly into currency minor units, the item price and event value should use one consistent rounding policy.

### Source locations

- `includes/class-wc-abstract-google-analytics-js.php`
  - `get_formatted_product()`
  - `get_formatted_order()`
- `assets/js/src/utils/index.js`
  - `getProductFieldObject()`

### Additional context

GA4 event reference:

<https://developers.google.com/analytics/devguides/collection/ga4/reference/events#purchase>

Related upstream work:

- Issue #417 reported that purchase items ignored order-level coupon discounts:
  <https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/417>
- PR #419 introduced the current comparison between the product price and
  `WC_Order_Item_Product::get_total()`:
  <https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/419>

This report is a follow-up to that completed work: the new discount handling
uses mismatched tax and quantity bases and therefore produces false or
inflated discounts.

The current trunk implementation still contains the same item-price comparison as version 2.3.0.

### LINKED ISSUE #642: purchase.value includes tax and shipping instead of the GA4 items value
### Describe the bug:

The `purchase()` formatter uses `order.totals.total_price` as event-level `value`. `get_formatted_order()` populates that field from `WC_Order::get_total()`, which includes tax and shipping.

The GA4 event reference defines `value` as the sum of `price × quantity` for all items and says not to include tax or shipping. The integration already follows this definition for checkout events by summing tax-exclusive item line totals, but `purchase` uses a different gross-order-total basis.

### Steps to reproduce:

1. Configure WooCommerce to enter prices including tax.
2. Purchase one SEK 95 product that includes SEK 5.38 tax, with zero shipping.
3. Inspect the `purchase` hit in Tag Assistant.

### Expected behavior:

```text
price: 89.62
quantity: 1
value: 89.62
tax: 5.38
shipping: 0
```

Event-level `value` should equal the emitted items value and exclude tax and shipping.

### Actual behavior:

```text
value: 95
tax: 5.38
shipping: 0
```

A second order with quantity 1 and a 10% coupon produced:

```text
price: 80.66
quantity: 1
value: 85.50
tax: 4.84
shipping: 0
```

The expected `value` for that emitted item is 80.66.

### Additional details:

Affected version: Google Analytics for WooCommerce 2.3.0.

Source locations:

- `includes/class-wc-abstract-google-analytics-js.php`
  - `get_formatted_order()` populates `total_price` from `WC_Order::get_total()`.
- `assets/js/src/tracker/data-formatting.js`
  - `purchase()` formats `order.totals.total_price` as `value`.
  - `getCheckoutData()` already documents and implements the tax- and shipping-exclusive GA4 definition for checkout events.

GA4 event reference:

<https://developers.google.com/analytics/devguides/collection/ga4/reference/events#purchase>

The current trunk implementation still contains the same purchase-value calculation as version 2.3.0.

costasovo:

This PR fixes two related problems in the data the plugin sends to Google Analytics for the `purchase` event.

#### Issue 1 — false or inflated item discounts (#641)

To report a discount on a purchased item, the plugin compared two prices: the product's catalog price and the amount actually paid for the order line. Those two numbers were not comparable — the catalog price is **per unit** and, in stores that enter prices *including* tax, it **includes tax**; the paid amount is the **whole line** (all units) **without tax**. Any difference between them was reported to GA as a "discount".

In a store with a product priced 95 including 6% tax, this is what GA received:

| Order | What GA received | What is wrong |
|---|---|---|
| 1 unit, no coupon | `price: 89.62`, **`discount: 5.38`** | There was no discount — the 5.38 is the tax. |
| 2 units, no coupon | **`price: 95`**, no discount | Price is now the tax-inclusive catalog price, because the two-unit line total happens to be larger than the single-unit catalog price. |
| 1 unit, 10% coupon | `price: 80.66`, **`discount: 14.34`** | The real discount is 8.96; the reported one is the discount plus the tax. |

Stores that enter prices *excluding* tax were affected too, just less visibly: with a coupon and more than one unit, the discount was silently dropped (`price: 95`, no discount, instead of `price: 85.50`, `discount: 9.50`).

#### Issue 2 — purchase revenue included tax and shipping (#642)

The event-level `value` (the revenue GA attributes to the purchase) was taken from the WooCommerce order total, which includes tax and shipping. Google's GA4 reference defines `value` as the sum of item price × quantity, explicitly **excluding** tax and shipping — and the plugin already followed that definition for the checkout events (`begin_checkout`, `add_shipping_info`, `add_payment_info`), so `purchase` disagreed with the rest of the funnel.

For the same 95 product (1 unit, no coupon), GA received `value: 95` alongside `tax: 5.38` — the tax counted twice; the expected value is 89.62.

#### The fix

Every amount for a purchased item is now taken from the **order line** rather than the catalog, and expressed on one consistent basis: **per unit, excluding tax**. The pre-coupon and post-coupon unit prices then compare like for like, so `discount` is reported only when a coupon actually reduced the price, and `price` is always the tax-exclusive unit price the customer paid.

The purchase `value` is calculated from those item prices (price × quantity, summed), so it excludes tax and shipping as GA4 expects and matches the checkout events.

Result for the three orders above: `price: 89.62` with no discount / `value: 89.62`; `price: 89.62`, qty 2 / `value: 179.24`; `price: 80.66`, `discount: 8.96` / `value: 80.66`.

Closes #641, STORMA-174
Closes #642, STORMA-175

#### Screenshots

|Usecase|Before|After|
|---|---|---|
|1 unit, no coupon|before-case1-qty1-2-thankyou|case1-qty1-2-thankyou|
|2 units, no coupon|before-case2-qty2-2-thankyou|case2-qty2-2-thankyou|
|1 unit, 10% coupon|before-case3-qty1-coupon-2-thankyou|case3-qty1-coupon-2-thankyou|

#### Things reviewers should be aware of

* **Reported GA revenue changes.** `value` drops from the gross order total to net item revenue (no tax, no shipping). This is what GA4 prescribes, but stores comparing GA revenue with WooCommerce reports will see the difference.
* **Order fees are not part of `value`.** Payment surcharges, or plugins that implement discounts as negative fees instead of coupons, are no longer reflected in `value`. GA4 has no concept of fees, and the checkout events already excluded them. Coupon discounts are unaffected — WooCommerce spreads them into the line totals we now read.

### Checks:

Setup: enable taxes, set WooCommerce > Settings > Tax to "Yes, I will enter prices inclusive of tax", add a standard tax rate (e.g. 6%), create a simple product priced 95, a 10% coupon, and make sure a zero-cost shipping method and a payment method are available. Inspect the `purchase` event with Google Tag Assistant, or by reading `window.dataLayer` on the order-received page.

1. Buy 1 unit without a coupon → item `price: 89.62`, **no** `discount`; event `value: 89.62`, `tax: 5.38`.
2. Buy 1 unit with the 10% coupon → item `price: 80.66`, `discount: 8.96`; event `value: 80.66`, `tax: 4.84`.
3. Buy 2 units without a coupon → item `price: 89.62`, `quantity: 2`, no `discount`; event `value: 179.24` (not the tax-inclusive `95` per unit that 2.3.0 sent).
4. Switch to prices *excluding* tax (or disable taxes) and buy 2 units with the 10% coupon → item `price: 85.50`, `discount: 9.50`; event `value: 171`.

Automated coverage: three new PHPUnit tests for the per-unit, tax-exclusive item amounts, plus updated and new E2E assertions checking that `value` excludes tax and shipping and uses the discounted item price.

### Changelog entry

> Fix - Purchase event no longer reports a false item discount in stores that enter prices inclusive of tax, and no longer drops coupon discounts for quantities above one; item price and discount are now per unit and tax-exclusive.
> Fix - Purchase event value now sums the item prices × quantity, excluding tax and shipping as GA4 expects, instead of the gross order total.

No issues found.

### DIFF
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -207,7 +207,7 @@ export const purchase = ( { order } ) => {
 		return false;
 	}
 
-	const items = order.items.map( getProductFieldObject );
+	const items = order.items.map( ( item ) => getProductFieldObject( item ) );
 	const minorUnit = order.totals.currency_minor_unit;
 
 	return {
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -207,23 +207,27 @@ export const purchase = ( { order } ) => {
 		return false;
 	}
 
+	const items = order.items.map( getProductFieldObject );
+	const minorUnit = order.totals.currency_minor_unit;
+
 	return {
 		transaction_id: order.id,
 		affiliation: order.affiliation,
 		currency: order.totals.currency_code,
+		// GA4 defines purchase value as the sum of price × quantity for the
+		// items, excluding tax and shipping.
+		// https://developers.google.com/analytics/devguides/collection/ga4/reference/events#purchase
 		value: formatPrice(
-			order.totals.total_price,
-			order.totals.currency_minor_unit
+			items.reduce(
+				( sum, { price = 0, quantity = 1 } ) =>
+					sum + Math.round( price * 10 ** minorUnit ) * quantity,
+				0
+			),
+			minorUnit
 		),
-		tax: formatPrice(
-			order.totals.tax_total,
-			order.totals.currency_minor_unit
-		),
-		shipping: formatPrice(
-			order.totals.shipping_total,
-			order.totals.currency_minor_unit
-		),
-		items: order.items.map( getProductFieldObject ),
+		tax: formatPrice( order.totals.tax_total, minorUnit ),
+		shipping: formatPrice( order.totals.shipping_total, minorUnit ),
+		items,
 	};
 };
 
--- a/assets/js/src/tracker/data-formatting.js
+++ b/assets/js/src/tracker/data-formatting.js
@@ -219,8 +219,12 @@ export const purchase = ( { order } ) => {
 		// https://developers.google.com/analytics/devguides/collection/ga4/reference/events#purchase
 		value: formatPrice(
 			items.reduce(
+				// Round each line to whole minor units: quantity may be decimal.
 				( sum, { price = 0, quantity = 1 } ) =>
-					sum + Math.round( price * 10 ** minorUnit ) * quantity,
+					sum +
+					Math.round(
+						Math.round( price * 10 ** minorUnit ) * quantity
+					),
 				0
 			),
 			minorUnit
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -577,13 +577,22 @@ abstract class WC_Abstract_Google_Analytics_JS {
 				continue;
 			}
 
+			// Both amounts are per-unit and tax-exclusive so the client-side
+			// discount calculation compares values on the same basis. The
+			// catalog price from get_formatted_product() is not suitable here:
+			// it includes tax when the store enters prices inclusive of tax.
+			// A zero-quantity line has a zero total, so dividing by 1 keeps it at 0.
+			$unit_divisor = max( 1, (int) $item->get_quantity() );
+
 			$items[] = array_merge(
 				$this->get_formatted_product( $product ),
 				array(
 					'quantity'                    => $item->get_quantity(),
-					// The method get_total() will return the price after coupon discounts.
-					// https://github.com/woocommerce/woocommerce/blob/54eba223b8dec015c91a13423f9eced09e96f399/plugins/woocommerce/includes/class-wc-order-item-product.php#L308-L310
-					'price_after_coupon_discount' => $this->get_formatted_price( $item->get_total() ),
+					'prices'                      => array(
+						'price'               => $this->get_formatted_price( (float) $item->get_subtotal() / $unit_divisor ),
+						'currency_minor_unit' => wc_get_price_decimals(),
+					),
+					'price_after_coupon_discount' => $this->get_formatted_price( (float) $item->get_total() / $unit_divisor ),
 				)
 			);
 		}
--- a/includes/class-wc-abstract-google-analytics-js.php
+++ b/includes/class-wc-abstract-google-analytics-js.php
@@ -581,8 +581,10 @@ abstract class WC_Abstract_Google_Analytics_JS {
 			// discount calculation compares values on the same basis. The
 			// catalog price from get_formatted_product() is not suitable here:
 			// it includes tax when the store enters prices inclusive of tax.
-			// A zero-quantity line has a zero total, so dividing by 1 keeps it at 0.
-			$unit_divisor = max( 1, (int) $item->get_quantity() );
+			// Quantities may be decimal (woocommerce_stock_amount filter), so do not
+			// cast to int. A zero-quantity line has a zero total, so dividing by 1 keeps it at 0.
+			$quantity     = (float) $item->get_quantity();
+			$unit_divisor = $quantity > 0 ? $quantity : 1;
 
 			$items[] = array_merge(
 				$this->get_formatted_product( $product ),
--- a/tests/e2e/specs/gtag-events/blocks-pages.test.js
+++ b/tests/e2e/specs/gtag-events/blocks-pages.test.js
@@ -871,9 +871,7 @@ test.describe( 'GTag events on block pages', () => {
 			// excluding tax and shipping.
 			const itemsTotal = simpleProductPrice + simpleProductPrice + 18.99;
 			expect( data.cu ).toEqual( 'USD' );
-			expect( data[ 'epn.value' ] ).toEqual(
-				itemsTotal.toFixed( 2 ).toString()
-			);
+			expect( data[ 'epn.value' ] ).toEqual( itemsTotal.toFixed( 2 ) );
 			expect( data[ 'epn.tax' ] ).toEqual( '0' );
 			expect( data[ 'epn.shipping' ] ).toEqual( shipping.toString() );
 		} );
--- a/tests/e2e/specs/gtag-events/blocks-pages.test.js
+++ b/tests/e2e/specs/gtag-events/blocks-pages.test.js
@@ -867,11 +867,12 @@ test.describe( 'GTag events on block pages', () => {
 			);
 
 			const shipping = 10;
-			const total =
-				simpleProductPrice + simpleProductPrice + 18.99 + shipping;
+			// GA4 purchase value is the sum of item price × quantity,
+			// excluding tax and shipping.
+			const itemsTotal = simpleProductPrice + simpleProductPrice + 18.99;
 			expect( data.cu ).toEqual( 'USD' );
 			expect( data[ 'epn.value' ] ).toEqual(
-				total.toFixed( 2 ).toString()
+				itemsTotal.toFixed( 2 ).toString()
 			);
 			expect( data[ 'epn.tax' ] ).toEqual( '0' );
 			expect( data[ 'epn.shipping' ] ).toEqual( shipping.toString() );
--- a/tests/e2e/specs/gtag-events/classic-pages.test.js
+++ b/tests/e2e/specs/gtag-events/classic-pages.test.js
@@ -903,6 +903,47 @@ test.describe( 'GTag events on classic pages', () => {
 		}
 	} );
 
+	test( 'Purchase event reports no discount when prices are entered inclusive of tax', async ( {
+		page,
+	} ) => {
+		await setOptions( {
+			woocommerce_calc_taxes: 'yes',
+			woocommerce_prices_include_tax: 'yes',
+			woocommerce_tax_based_on: 'billing',
+		} );
+		const taxRateID = await createCaliforniaTaxRate();
+
+		try {
+			await simpleProductAddToCart( page, simpleProductID );
+
+			const event = trackGtagEvent( page, 'purchase', 'checkout' );
+			await checkout( page );
+
+			await event.then( ( request ) => {
+				const data = getEventData( request, 'purchase' );
+				// The item price is the tax-exclusive unit price (9.99 incl. 10%
+				// tax), and there is no coupon, so no `ds` (discount) field may be
+				// present — toEqual asserts the exact key set.
+				expect( data.product1 ).toEqual( {
+					id: simpleProductID.toString(),
+					nm: 'Simple product',
+					ca: 'Uncategorized',
+					qt: '1',
+					pr: '9.08',
+				} );
+				expect( data[ 'epn.tax' ] ).toEqual( '0.91' );
+				expect( data[ 'epn.shipping' ] ).toEqual( '10' );
+				expect( data[ 'epn.value' ] ).toEqual( '9.08' );
+			} );
+		} finally {
+			await deleteTaxRate( taxRateID );
+			await setOptions( {
+				woocommerce_calc_taxes: 'no',
+				woocommerce_prices_include_tax: 'no',
+			} );
+		}
+	} );
+
 	test( 'Begin checkout and purchase events use discounted item price', async ( {
 		page,
 	} ) => {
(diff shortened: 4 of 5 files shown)


includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -577,13 +577,21 @@ public function get_formatted_order( $order ): array {
...
 				array(
 					'quantity'                    => $item->get_quantity(),
-					// The method get_total() will return the price after coupon discounts.
-					// https://github.com/woocommerce/woocommerce/blob/54eba223b8dec015c91a13423f9eced09e96f399/plugins/woocommerce/includes/class-wc-order-item-product.php#L308-L310
-					'price_after_coupon_discount' => $this->get_formatted_price( $item->get_total() ),
+					'prices'                      => array(
+						'price'               => $this->get_formatted_price( (float) $item->get_subtotal() / $quantity ),
+						'currency_minor_unit' => wc_get_price_decimals(),
+					),
+					'price_after_coupon_discount' => $this->get_formatted_price( (float) $item->get_total() / $quantity ),
```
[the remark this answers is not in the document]
costasovo: `get_quantity()` always returns an int, so the non-numeric case cannot occur. For a zero-quantity line (only reachable by editing an order in admin) WooCommerce sets the line total to 0, so the emitted item is `price: 0`, `quantity: 0` and contributes 0 to `value` — that is the correct revenue for a line nobody paid for, not under-reporting; reporting `quantity: 1` for it would be wrong. The `max( 1, … )` is purely a division-by-zero guard, so I renamed it to `$unit_divisor` and documented the zero-quantity behaviour to make that explicit (3ab365e).

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -577,13 +577,21 @@ public function get_formatted_order( $order ): array {
...
 				)
```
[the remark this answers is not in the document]
costasovo: Keeping `array()` here on purpose: this file uses `array()` throughout (42 occurrences, no `[]`), and this hunk is nested inside an existing `array_merge( …, array(` call, so mixing the two syntaxes in one expression would read worse than following the file convention. PHPCS passes. A file-wide switch to short syntax is a reasonable separate cleanup.

assets/js/src/tracker/data-formatting.js
```diff
@@ -207,23 +207,27 @@ export const purchase = ( { order } ) => {
 		return false;
 	}
 
+	const items = order.items.map( getProductFieldObject );
```
eason9487: `map()` calls its callback with `(element, index, array)`, so `getProductFieldObject()` receives the array index as its `quantity` argument here. Nothing goes wrong today, but would it be worth passing the item explicitly anyway, so the two never depend on each other?

includes/class-wc-abstract-google-analytics-js.php
```diff
@@ -577,13 +577,22 @@ public function get_formatted_order( $order ): array {
 				continue;
 			}
 
+			// Both amounts are per-unit and tax-exclusive so the client-side
+			// discount calculation compares values on the same basis. The
+			// catalog price from get_formatted_product() is not suitable here:
+			// it includes tax when the store enters prices inclusive of tax.
+			// A zero-quantity line has a zero total, so dividing by 1 keeps it at 0.
+			$unit_divisor = max( 1, (int) $item->get_quantity() );
```
eason9487: Since what is needed here is a guard against dividing by zero, might it be better to check the quantity against `0.0` directly rather than casting it to `int`?

Asking because core moved in the other direction. woocommerce/woocommerce#58693 removed forced `int` casting on quantities so that extensions can use decimals. Its test plan covers placing an order with a decimal quantity and confirming it survives through to the order confirmation page, which is where this method runs.

tests/e2e/specs/gtag-events/classic-pages.test.js
eason9487, line 1: Would it be worth adding a case here that runs with `woocommerce_prices_include_tax: 'yes'` and asserts that the item carries no `discount`?

The reason for asking is that the new PHPUnit tests pin down the values behind the false discount, but nothing asserts the `ds` field they would produce, and no case runs with prices entered inclusive of tax. `woocommerce_prices_include_tax` only ever appears set to `'no'`, and the single `ds` assertion lives in the coupon test where a discount is expected.

tests/e2e/specs/gtag-events/blocks-pages.test.js
```diff
@@ -867,11 +867,12 @@ test.describe( 'GTag events on block pages', () => {
...
 			const shipping = 10;
-			const total =
-				simpleProductPrice + simpleProductPrice + 18.99 + shipping;
+			// GA4 purchase value is the sum of item price × quantity,
+			// excluding tax and shipping.
+			const itemsTotal = simpleProductPrice + simpleProductPrice + 18.99;
 			expect( data.cu ).toEqual( 'USD' );
 			expect( data[ 'epn.value' ] ).toEqual(
-				total.toFixed( 2 ).toString()
+				itemsTotal.toFixed( 2 ).toString()
```
eason9487: The `.toString()` here is pre-existing, not something this PR introduced. `toFixed()` already returns a string, and since the expression is being rewritten anyway, maybe it's worth dropping it in passing.

tests/e2e/specs/gtag-events/classic-pages.test.js
```diff
@@ -855,11 +855,12 @@ test.describe( 'GTag events on classic pages', () => {
...
 			const shipping = 10;
-			const total =
-				simpleProductPrice + simpleProductPrice + 18.99 + shipping;
+			// GA4 purchase value is the sum of item price × quantity,
+			// excluding tax and shipping.
+			const itemsTotal = simpleProductPrice + simpleProductPrice + 18.99;
 			expect( data.cu ).toEqual( 'USD' );
 			expect( data[ 'epn.value' ] ).toEqual(
-				total.toFixed( 2 ).toString()
+				itemsTotal.toFixed( 2 ).toString()
```
eason9487: Same as https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650#discussion_r3811467856.

eason9487 (commented):
Thanks for the fix! Tested locally with prices entered inclusive of tax, and it works as described.

The one thing I would like to settle before approving is whether the quantity should be cast to an integer. The other comments are not blockers.

assets/js/src/tracker/data-formatting.js
costasovo: Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650/commits/2840f900a24b98654a339a4d3d4eb71e7a692efa

includes/class-wc-abstract-google-analytics-js.php
costasovo: Great catch! I was not aware that Woo supports decimal quantities. Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650/commits/d2b6cf1950a00e876b55aefc4c362c8d10812e5f and in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650/commits/2aac78da515d49fadaf9d33b63fa9d17d2419bb6 ([Issue flagged by Code Rabbit](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650#pullrequestreview-4972212721) - without rounding the price might not match)

When working on the fix, CC flagged that the code for tracking cart events has the same problem. The fix is not trivial (more fixes in the JS part), so I didn't want to fold it into this PR, and I created a follow-up STORMA-177.

tests/e2e/specs/gtag-events/blocks-pages.test.js
costasovo: Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650/commits/2840f900a24b98654a339a4d3d4eb71e7a692efa

tests/e2e/specs/gtag-events/classic-pages.test.js
costasovo: Fixed in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650/commits/2840f900a24b98654a339a4d3d4eb71e7a692efa

tests/e2e/specs/gtag-events/classic-pages.test.js
costasovo: Added E2E test in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/650/commits/c16a20c9a6fc8eae68d9d3800b99c633ef051510

costasovo:
@eason9487 Thank you for the detailed review! I added four new commits with fixes and the E2E test.

eason9487 (approved):
Thanks for the update! LGTM.

PR #662, opened 2026-09-07
Strip sensitive query parameters from the URLs reported to Google Analytics

pavel-mailpoet:

Closes [STORMA-180](https://linear.app/a8c/issue/STORMA-180/sec-1-redact-sensitive-url-params-before-analytics-dispatch-2-1).

The order-received URL carries the order key (`?key=wc_order_…`), a bearer token for the order. gtag sent it to Google in `dl` with every hit and in `dr` on the next page, where anyone with access to the GA4 property could replay it.

The page URL and referrer are now filtered in the browser before gtag reads them: a filterable denylist plus a value rule on every page, and an allowlist on the order-received and order-pay pages. The value rule matches `wc_` rather than `wc_order_`, since WooCommerce only guarantees that half of the prefix, and matches anywhere in the value so a key nested in a return URL is caught too. Filtering runs client-side because the snippet is built before the main query is parsed and full-page caches would otherwise serve one URL for every query string. The helper is registered outside the `woocommerce_gtag_snippet` filter, so replacing the snippet does not silently drop the protection. URLs with nothing to strip pass through untouched, and `page_location`/`page_referrer` set through `woocommerce_ga_gtag_config` win.

On a site the WP Consent API reports as opt-in, a visitor who has not answered the banner is now sent an explicit `denied` instead of being left on the regional defaults. Sites reported as opt-out, and sites where nothing declares a consent type, keep the defaults, so they do not lose measurement they have today.

Not covered: browsers still send the page URL in the `Referer` header of the requests the page makes, and other tags on the page report their own `page_location`.

### Checks:

### Screenshots:

n/a

1. Run `npm run test:js`, `npm run test:php` and `npm run test:e2e`.
2. Place an order. In DevTools → Network, filter `collect`: `dl` on the `page_view` and `purchase` hits has no `key=`. Go to the shop: `dr` has no `key=`.
3. Open `/shop/?utm_source=x&foo=bar`: `dl` matches the browser URL and the `config` entry in `window.dataLayer` has no `page_location`.

### Additional details:

New filters: `woocommerce_ga_redacted_url_params`, `woocommerce_ga_order_page_allowed_url_params`, `woocommerce_ga_url_redaction_enabled`, documented in the README along with `wcGoogleAnalyticsIntegration.redactGtagConfig()`, which a replacement snippet can call to keep the redaction.

The redaction helper is JavaScript embedded in a PHP string, so `npm run test:js` reads it out of the source file and runs its branches on the node test runner. The E2E helper now compares hostname and path of `dl`, because an explicit `page_location` keeps the port that gtag's default omits.

### Changelog entry

> Fix - Strip the order key and other sensitive query parameters from the page URL and referrer sent to Google Analytics, and send an explicit denied consent state for undecided visitors on opt-in sites.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

No issues found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,10 @@ The extension does not provide any UI, like a cookie banner, to let your visitor
 
 Each of those extensions may require additional setup or registration. Usually, the basic default setup works out of the box, but there may be some integration caveats. Here are a couple of the most frequent ones:
 
+##### Visitors who have not answered the banner yet
+
+When the WP Consent API reports that the site is opt-in, a visitor who has not answered the banner is reported to Google as `denied` for every consent type, rather than being left on the regional defaults. Sites the WP Consent API reports as opt-out, and sites where no extension declares a consent type at all, are unaffected: their visitors keep the defaults until they make a choice.
+
 ##### GA4W overwrites the consent mode defaults set by the other extension
 
 If the additional extension you chose sets its own default state of consent modes, different than the one we set, and you would like to make sure we'll not overwrite that, you can use the `woocommerce_ga_gtag_consent_modes` snippet to change or disable our setup:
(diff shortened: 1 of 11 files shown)


triple0t (approved):
LGTM

PR #664, opened 2026-09-08
Harden the release and test pipelines

pavel-mailpoet:

Closes STORMA-181 (findings #3, #0, #4 from the v2.4.0 security triage). CI/CD only, no change to installed sites.

**Deploy workflow (findings #3 and #0)**
- Split `deploy.yml` into `validate`, `dry-run` and `deploy` jobs. Validation refuses any ref that is not `release/<version>` or `hotfix/<version>` matching the version input, and records the commit in the step summary.
- Production runs use the protected `production` environment and refuse to start if the branch moved since the approved commit.
- Dry runs need no approval and never receive the WordPress.org SVN password. Partner credentials stay so QIT still runs.
- The private `woo-product-deploy` action is pinned to a commit SHA and checked out without persisted credentials. The unused `workflow_call` trigger is removed.
- Concurrency group per ref, a descriptive run name, and the version pattern aligned with the action's own check.

**Test setup (finding #4)**
- Removed the unpinned `db.php` download from `markoheijnen/wp-mysqli@master`. The file only defined `WP_USE_EXT_MYSQL` as false, which WordPress ignores on PHP 7+.
- The PHP unit test workflow now runs when `bin/*.sh` changes, and the version check is fetched over HTTPS with curl failing on errors.

**Same patterns found in the sweep**
- `.wp-env.test.json` pinned WP‑API/Basic‑Auth to a commit instead of `master.zip`.
- Hook documentation output, PHP unit test matrix values and the release PR URL are read from env instead of being expanded into shell scripts.

**Repo settings done separately (all verified):** the `production` environment exists with the `ballade` team as required reviewer, self‑review prevented, admin bypass off, and `release/*` plus `hotfix/*` branch policies. The WP.org SVN secrets moved from repo secrets into that environment. A ruleset blocks force pushes and deletion on `release/*` and `hotfix/*`. Unused `QIT_PARTNER_*` repo secrets removed.

**Upstream follow‑up:** woocommerce/woo-product-deploy#7 pins the `qit-action` checkout, stops persisting the PAT in the target checkout, and adds a `commit` input so the approved commit is the one woorelease builds. Until it merges, dry runs still expose the org PAT to code on the branch. After it merges, a small follow‑up here bumps `WOO_PRODUCT_DEPLOY_SHA` and passes `commit: ${{ needs.validate.outputs.sha }}` in the deploy job.

### Checks:

### Screenshots:

_N/A_

1. Confirm the `PHP Unit Tests` and `E2E Tests` checks run on this PR and pass. The unit test job exercises the install script without `db.php`; the E2E job installs Basic‑Auth from the pinned zip.
2. Negative test of the gate, from this branch: run the Deploy Product workflow with `version=2.4.1` and dry run on. Expected: `Validate release ref` fails with "Refusing to deploy", the other jobs are skipped, no approval request appears.
3. Before the next release, run a dry run from a real `release/x.y.z` branch and confirm the run title shows "(dry run)", the action checkout uses the pinned SHA, and QIT runs.
4. On the first production run, confirm the `ballade` approval prompt appears and the summary of `Validate release ref` shows the commit that will ship.

### Additional details:

Removing `db.php` is stricter than the pin the issue asked for: the drop‑in was dead code for every PHP and WordPress version in the matrix, so there is nothing left to pin.

### Changelog entry

> Dev - Harden the release and test pipelines: pin the deploy action and test dependencies, gate production deploys behind a protected environment.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,14 +1,17 @@
 name: Deploy Product
 
 # **What it does**: Builds and publishes a release to WordPress.org, Woo.com and GitHub.
-# **How to use it**: Run it from the `release/x.y.z` branch (or the `x.y.z` tag) that
-# matches the `version` input. Dry runs need no approval and never see the WP.org
-# publishing password. Production runs wait for approval on the `production` environment
-# and ship the exact commit that was approved.
+# **How to use it**: Run it from the `release/x.y.z` or `hotfix/x.y.z` branch that matches
+# the `version` input. Dry runs need no approval and never see the WP.org publishing
+# password. Production runs wait for approval on the `production` environment and refuse
+# to start if the branch moved since the approved commit. The action itself resolves the
+# branch by name, so the branch must also be protected against force pushes.
 #
 # The `woocommerce/woo-product-deploy` action is pinned to a commit SHA. Dependabot does
 # not track `repository:` refs in `actions/checkout`, so bump `WOO_PRODUCT_DEPLOY_SHA`
 # through a reviewed PR after checking the upstream diff.
+
+run-name: Deploy ${{ inputs.version }} from ${{ github.ref_name }}${{ inputs.simulate && ' (dry run)' || '' }}
 
 on:
   workflow_dispatch:
@@ -41,6 +44,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
 env:
   # woocommerce/woo-product-deploy trunk @ 2026-08-31
   WOO_PRODUCT_DEPLOY_SHA: 5f5b49d49b8a3681a7a99efec089187666885759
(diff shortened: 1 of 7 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,5 +1,18 @@
 name: Deploy Product
 
+# **What it does**: Builds and publishes a release to WordPress.org, Woo.com and GitHub.
```
eason9487: Two related observations now that the header documents the publish targets.

1. This workflow publishes only to WordPress.org and GitHub. WooCommerce.com is reached by a separate mechanism outside it, so listing Woo.com here could read as if this run pushes there too. If the summary keeps it, the name would be "WooCommerce.com" rather than "Woo.com".
2. For a WordPress.org-hosted plugin like this one, the `woo-product-deploy` action does not run its WooCommerce.com deploy step, so the `deploy_woo` and `wccom_product_id` inputs have no effect on a run here. Would it be worth removing those two inputs and hardcoding `wccom_release` as part of this PR, so operators are not presented with options that do nothing?

.github/workflows/deploy.yml
```diff
@@ -27,64 +40,146 @@ on:
...
+        env:
+          VERSION: ${{ inputs.version }}
+          REF: ${{ github.ref }}
+          REF_NAME: ${{ github.ref_name }}
+          SHA: ${{ github.sha }}
+          SIMULATE: ${{ inputs.simulate }}
+        run: |
+          # Same pattern as validate-inputs.sh in woo-product-deploy, so a version that
+          # passes here also passes inside the action.
+          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9]+)?$ ]]; then
```
eason9487: A leading zero passes the version check here, e.g. `01.2.3` or `2.004.2`, and the same pattern is in `prepare-release-via-deploy.yml`. Since the value is typed by hand on dispatch, would it be worth rejecting a leading zero?

.github/workflows/scripts/create-release-pr.mjs
```diff
@@ -22,6 +22,7 @@ export default async ( { context, github, version } ) => {
    woorelease cl:generate --release --product_version=${ version } ${ repoURL }
    \`\`\`
 1. [ ] Automated tests are passing.
+1. [ ] Wait for the automated hooks documentation commit on this branch (it runs after the changelog changes). A production run fails if the branch moves after approval.
```
eason9487: For someone running the release for the first time, "the changelog changes" might be a bit ambiguous: a few steps earlier the checklist uses "changelog entries" for `readme.txt`, so which changelog this step waits on is not self-evident. It might be clearer to name the file here.

eason9487 (approved):
Thanks for the work! The inline comments are all minor and non-blocking.

pavel-mailpoet:
Thanks @eason9487, all three are addressed.

**Publish targets and the WooCommerce.com inputs** – both fixed. The header now says WordPress.org and GitHub only, with a line explaining that the action skips its WooCommerce.com step for a WP.org-hosted product. 

**Leading zeros** – rejected now, in `deploy.yml` and in `prepare-release-via-deploy.yml`. 

**Release checklist wording** – the step now names the file: it waits on the `changelog.txt` commit that `cl:generate` makes, which is what the hook documentation workflow triggers on.

Worth noting for merge order: `WOO_PRODUCT_DEPLOY_SHA` here pins the current tip of `woo-product-deploy` trunk, which is the commit woocommerce/woo-product-deploy#7 branches from, and this PR never passes `commit:`. So the two are independent, and this one can merge first. The follow-up to bump the SHA and pass the approved commit stays as described in the description.

